PR #6, opened 2025-08-24
Corrects checkout and cloning logic 

roshkhatri:
This corrects checkout and cloning logic while benchmarking a PR and otherwise when we provide the valkey-path.

This also adds a `cluster reset hard` for cluster mode enabled as while running it id the cluster is already setup the cluster setup would fail.

### DIFF
--- a/benchmark.py
+++ b/benchmark.py
@@ -314,9 +314,9 @@ def run_benchmark_matrix(
 
         if not args.use_running_server:
             if args.valkey_path:
-                builder.terminate_server()
+                builder.terminate_valkey()
             else:
-                builder.cleanup_terminate()
+                builder.terminate_and_clean_valkey()
 
 
 # ---------- Entry point ------------------------------------------------------
--- a/benchmark.py
+++ b/benchmark.py
@@ -255,8 +255,6 @@ def run_benchmark_matrix(
         parse_core_range(args.client_cpu_range)
         bench_core_range = args.client_cpu_range
 
-    cleanup_required = not args.valkey_path or not args.use_running_server
-
     valkey_dir = (
         Path(args.valkey_path) if args.valkey_path else Path(f"../valkey_{commit_id}")
     )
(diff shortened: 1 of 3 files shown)


benchmark.py
```diff
@@ -314,8 +312,11 @@ def run_benchmark_matrix(
         except Exception as exc:
             logging.warning(f"Failed to update completed_commits.json: {exc}")
 
-        if cleanup_required:
-            builder.cleanup_terminate()
+        if not args.use_running_server:
+            if args.valkey_path:
+                builder.terminate_server()
+            else:
+                builder.cleanup_terminate()
```
sarthakaggarwal97, line 319: what is the difference between the two functions?

sarthakaggarwal97: I see `terminate` is being called inside `cleanup_terminate`. I suggested better naming below to avoid any confusion.

valkey_build.py
```diff
@@ -34,21 +34,21 @@ def _run(self, command: Iterable[str], cwd: Optional[Path] = None) -> None:
...
-                f"Using existing Valkey repository at {self.valkey_dir} - skipping clone and checkout"
-            )
-            return
-
-        # Only clone if directory doesn't exist
-        logging.info(f"Cloning Valkey repo into {self.valkey_dir}...")
-        self._run(["git", "clone", self.repo_url, str(self.valkey_dir)])
+            logging.info(f"Using existing Valkey repository at {self.valkey_dir}")
+        else:
+            # Only clone if directory doesn't exist
```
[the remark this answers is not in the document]
roshkhatri: ># If valkey_path exists, assume it has all the commits we need and skip cloning

># Only clone if directory doesn't exist

roshkhatri:
Yes it fixes this as we were cleaning up the code that we checkout in the PR. Later that context wont be available and that is why we were seeing the error.

PR #7, opened 2025-09-03
Change read commands for SET and ZSET

ranshid:
Currently we use SPOP and ZPOPMIN as the test scenarios for sets and sorted sets This is not good as they are not read commands and will consume the sets while we would like to have a consistent benckmark.

Depending on https://github.com/valkey-io/valkey/pull/2575 will be merged, we can fix that and support better candidates for SET and ZSET read commands benchmarking


rainsupreme (approved):
LGTM!

PR #8, opened 2025-09-04
Adds io-thread support

roshkhatri:


SoftlyRaining (approved):

roshkhatri:
`--server-cpu-range` already exists in the implementation. we need to set it, I think it should also be a part of the config file as we can change the server-cpu-ranges for each configs

Issue #24, opened 2026-01-04
Bug: Invalid metrics parsing for LRANGE


nesty92:
2026-01-04 03:41:09,006 [INFO] Benchmark output:
"test","rps","avg_latency_ms","min_latency_ms","p50_latency_ms","p95_latency_ms","p99_latency_ms","max_latency_ms"
"LPUSH (needed to benchmark LRANGE)","1079095.00","0.445","0.136","0.399","0.607","0.743","2.223"
"LRANGE_100 (first 100 elements)","249528.50","1.623","0.200","1.631","2.103","2.311","3.727"
"LRANGE_300 (first 300 elements)","85702.00","3.821","0.224","3.863","4.759","5.847","9.239"
"LRANGE_500 (first 500 elements)","49708.09","5.120","0.248","5.175","7.807","9.551","12.303"
"LRANGE_600 (first 600 elements)","39765.85","6.102","0.304","6.071","10.303","12.111","15.047"

2026-01-04 03:41:09,006 [INFO] Parsed metrics: {'timestamp': '2026-01-04T03:20:31+01:00', 'commit': 'perf/xrange', 'command': 'LRANGE', 'data_size': 16, 'pipeline': 10, 'clients': 50, 'rps': 1079095.0, 'avg_latency_ms': 0.445, 'min_latency_ms': 0.136, 'p50_latency_ms': 0.399, 'p95_latency_ms': 0.607, 'p99_latency_ms': 0.743, 'max_latency_ms': 2.223, 'cluster_mode': False, 'tls': False, 'duration': 2, 'benchmark_mode': 'duration', 'valkey_benchmark_threads': 2, 'warmup': 10, 'architecture': 'x86_64'}
2026-01-04 03:41:09,006 [INFO] Successfully wrote 1 metrics to results/perf/xrange/metrics.json

It's only parsing the first line and taking the result from the LPUSH

PR #25, opened 2026-01-04
fix: metrics parsing

nesty92:
- Add support for XRANGE
- Support multiples metrics like LRANGE, LRANGE_100

Close: https://github.com/valkey-io/valkey-perf-benchmark/issues/24


nesty92:
@roshkhatri fixed

PR #31, opened 2026-02-04
Adds custom server configs support

roshkhatri:
## Summary

Adds support for passing arbitrary configuration to `valkey-server` at startup via two new optional fields in `benchmark-configs.json`:

| Field | Purpose |
|---|---|
| `custom-server-configs` | Inline `{key: value}` map of `--key value` flags to pass through |
| `custom-server-config-file` | Path to a Valkey-format `.conf` file passed positionally |

Useful for setting workload-relevant tunables like `maxmemory`, `timeout`, `maxclients`, `tcp-keepalive`, `hz`, etc. without touching the harness.

### Example

```json
{
  "duration": 120,
  "commands": ["SET", "GET"],
  "cluster_mode": false,
  "tls_mode": false,
  "custom-server-config-file": "/etc/valkey/baseline.conf",
  "custom-server-configs": {
    "maxmemory": "4gb",
    "timeout": 300
  }
}
```

### Precedence

Configs are applied in this order on the `valkey-server` command line:

1. `custom-server-config-file` (positional, parsed first; lowest priority)
2. `custom-server-configs` (`--key value` flags)
3. Benchmark-managed defaults (highest priority via Valkey's last-wins semantics)

This means harness-critical settings (`port`, `daemonize`, `logfile`, `cluster-*`, etc.) always take effect regardless of what the user supplies.

### Validation

`validate_config` enforces:
- `custom-server-configs` is a dict
- Keys are strings
- Values are strings, ints, or floats. Booleans rejected because Python's `str(True)` does not produce a valid Valkey config value.
- `custom-server-config-file` is a string path

### Tests

- 20 validation tests in `tests/test_benchmark_config.py`
- 14 server-command tests in `tests/test_custom_server_configs.py` covering single-node, cluster mode, ordering, and defense-in-depth (benchmark defaults win on collision)

### DIFF
--- a/benchmark.py
+++ b/benchmark.py
@@ -53,6 +53,8 @@ OPTIONAL_CONF_KEYS = [
     "query_generation",
     "port",
     "module_startup_args",
+    "custom-server-configs",
+    "custom-server-config-file",
 ]
 
 
@@ -331,6 +333,22 @@ def validate_config(cfg: dict) -> None:
     if "port" in cfg:
         if not isinstance(cfg["port"], int) or cfg["port"] <= 0 or cfg["port"] > 65535:
             raise ValueError("'port' must be between 1 and 65535")
+    if "custom-server-configs" in cfg:
+        if not isinstance(cfg["custom-server-configs"], dict):
+            raise ValueError("'custom-server-configs' must be a dictionary")
+        for key, value in cfg["custom-server-configs"].items():
+            if not isinstance(key, str):
+                raise ValueError(
+                    f"'custom-server-configs' keys must be strings, got: {type(key)}"
+                )
+            # Note: bool is a subclass of int in Python, so check bool first.
+            if isinstance(value, bool) or not isinstance(value, (str, int, float)):
+                raise ValueError(
+                    f"'custom-server-configs' values must be strings or numbers, got: {type(value)}"
+                )
+    if "custom-server-config-file" in cfg:
+        if not isinstance(cfg["custom-server-config-file"], str):
+            raise ValueError("'custom-server-config-file' must be a string path")
 
     if "cluster_mode" in cfg and not isinstance(cfg["cluster_mode"], list):
         cfg["cluster_mode"] = parse_bool(cfg["cluster_mode"])
--- a/valkey_server.py
+++ b/valkey_server.py
@@ -136,6 +136,16 @@ class ServerLauncher:
 
         cmd.append(VALKEY_SERVER)
 
+        # Optional positional config file (must come right after the binary,
+        # before any --flag args). Subsequent --flag values override file values.
+        custom_conf_file = (
+            (self.config or {}).get("custom-server-config-file")
+            if hasattr(self, "config")
+            else None
+        )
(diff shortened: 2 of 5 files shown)


valkey_server.py
```diff
@@ -124,42 +124,73 @@ def _client_context(self, tls_mode: bool):
...
+            for key, value in custom_configs.items():
+                if key in configs:
+                    overridden.append(key)
+                configs[key] = str(value)
+
+            if overridden:
+                logging.warning(
+                    f"Custom configs overriding benchmark defaults: {', '.join(sorted(overridden))}"
+                )
+            logging.info(f"Applied custom server configs: {custom_configs}")
```
sarthakaggarwal97: Let's not log everything. It could have sensitive information as well.

sarthakaggarwal97 (commented):
I like the idea. I have provided some comments. 
Do we need to disable some important configurations like port, ip etc to avoid any side effects?

rainsupreme (commented):
I don't have anything to add beyond what Sarthak said. With the exception of a few tweaks it looks good to me!

valkey_server.py
sarthakaggarwal97: `requirepass`? I had this in mind. We will not use it but maybe someone else?

roshkhatri: removed the logging of the configs

benchmark.py
```diff
@@ -186,6 +187,20 @@ def validate_config(cfg: dict) -> None:
                     raise ValueError(
                         "'io-threads' must be a positive integer or list of positive integers"
                     )
+            # Validate optional custom-server-configs
+            elif k == "custom-server-configs":
+                if not isinstance(cfg["custom-server-configs"], dict):
+                    raise ValueError("'custom-server-configs' must be a dictionary")
```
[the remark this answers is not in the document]
roshkhatri: added another for `custom-server-config-file`

benchmark.py
```diff
@@ -331,6 +333,22 @@ def validate_config(cfg: dict) -> None:
     if "port" in cfg:
         if not isinstance(cfg["port"], int) or cfg["port"] <= 0 or cfg["port"] > 65535:
             raise ValueError("'port' must be between 1 and 65535")
+    if "custom-server-configs" in cfg:
```
sarthakaggarwal97, line 336: what happens if the configs are not valid? Is this something we should verify with `valkey.conf` file before attempting the benchmark? Trying to understand if there will be an additional churn.

sarthakaggarwal97 (commented):
Minor comment. Mostly looks good.

benchmark.py
roshkhatri: This is an additional churn, the conf should be valid or the process will fail from the server with the error. So we dont need to check the confs

sarthakaggarwal97 (approved):
Looks good to me!

benchmark.py
sarthakaggarwal97: okay since there is no additional cost, I think it's okay to proceed.

PR #50, opened 2026-03-19
Robust git ref resolution + integration tests

rainsupreme:
I got carried away and the original bug is already fixed, but I think the integration tests are nice, and this is a little more flexible.

This PR extracts git ref resolution into `utils/git_utils.py` with a multi-step
fallback that handles shallow clones and missing local refs — the common
case in CI environments. Replaces the HEAD-fallback workaround from #49
with proper ref resolution (local → fetch from origin → origin/ref →
FETCH_HEAD).

Adds an integration test suite (61 tests) covering the PR benchmark
comparison workflow, benchmark execution, git operations, and the new
ref resolution logic — all runnable without a live Valkey server.


roshkhatri (approved):
Overall looks good to me, great fall backs for different scenarios

PR #51, opened 2026-03-24
fix error margin to be consistent between measurement CI, PI, and delta between measurements

rainsupreme:
I made a mild mistake earlier - currently in the change % column the error margin is 1 standard deviation, though the report doesn't say this anywhere. Intuitively, it should be the same confidence level used in for the measurements it is comparing. (The change % error margin should generally be larger than the error margin for either of the two numbers being compared.)

Also, a 99% confidence interval is pretty aggressive. It seems that 95% is a more commonly used value for this kind of measurement, so with this change that's what we use throughout the report.

### DIFF
--- a/utils/compare_benchmark_results.py
+++ b/utils/compare_benchmark_results.py
@@ -12,11 +12,15 @@ import sys
 from typing import Dict, List, Tuple, Any, Optional
 from pathlib import Path
 
+import math
+
+from scipy import stats
+from uncertainties import ufloat
+
 # Optional dependencies for graphing functionality
 try:
     import matplotlib.pyplot as plt
     import numpy as np
-    from scipy import stats
     from matplotlib.ticker import FuncFormatter
 
     GRAPHING_AVAILABLE = True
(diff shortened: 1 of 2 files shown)


utils/compare_benchmark_results.py
```diff
@@ -12,24 +12,28 @@
 from typing import Dict, List, Tuple, Any, Optional
 from pathlib import Path
 
+import math
+
+from scipy import stats
+from uncertainties import ufloat
```
roshkhatri, line 18: Do we need to add any dependencies to the requirements files?

roshkhatri (commented):
Everything looks good to me, just one questions about dependency changes

utils/compare_benchmark_results.py
rainsupreme: they're already listed in requirements.in, no?

https://github.com/valkey-io/valkey-perf-benchmark/blob/d30b9baac656547bd91ae96d5293286576729b7e/requirements.in#L19

roshkhatri (approved):

PR #54, opened 2026-04-15
Removes  96 and adds 128 bytes for daily benchmark configs

roshkhatri:
Removes  96 and adds 128 bytes


sarthakaggarwal97 (approved):
Looks good to me!

PR #55, opened 2026-05-21
Record environment metadata in benchmark metrics

rainsupreme:
Add environment_metadata.py that captures system state alongside benchmark results:
- CPU model, governor, turbo/boost status, idle states, frequency
- ASLR status, THP mode, NUMA node count
- CPU pinning ranges (server + client)
- Benchmark tool version (git SHA from --version output)
- Whether --stabilize-environment was active

This enables reproducibility — reviewers can see exactly what environment produced a given result. Fields return 'not_available' gracefully on platforms that lack the relevant sysfs paths (e.g., Graviton has no governor/idle_states).

20 unit tests included.

### DIFF
--- a/environment_metadata.py
+++ b/environment_metadata.py
@@ -1,0 +1,198 @@
+"""Collect environment metadata for benchmark reproducibility.
+
+Records system state (CPU governor, turbo boost, kernel version, etc.)
+and benchmark tool version so results can be compared accurately across
+runs with different configurations.
+"""
+
+import logging
+import platform
+import subprocess
+from pathlib import Path
+from typing import Any, Dict, Optional
+
+
+def _run_cmd(cmd: str, default: str = "unknown") -> str:
+    """Run a shell command and return stripped stdout, or default on failure."""
+    try:
+        result = subprocess.run(
+            cmd, shell=True, capture_output=True, text=True, timeout=5
+        )
+        return result.stdout.strip() if result.returncode == 0 else default
+    except Exception:
+        return default
+
+
+def get_cpu_governor() -> str:
+    """Return the active CPU frequency governor, or 'not_available' for fixed-frequency CPUs."""
+    result = _run_cmd(
+        "cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_governor 2>/dev/null",
+        default="",
+    )
+    return result if result else "not_available"
+
+
+def get_turbo_boost_status() -> str:
+    """Return turbo/boost status: 'enabled', 'disabled', or 'not_available'."""
+    # Intel
+    intel_path = "/sys/devices/system/cpu/intel_pstate/no_turbo"
+    result = _run_cmd(f"cat {intel_path} 2>/dev/null", default="")
+    if result == "0":
+        return "enabled"
+    elif result == "1":
+        return "disabled"
+
+    # AMD
+    amd_path = "/sys/devices/system/cpu/cpufreq/boost"
+    result = _run_cmd(f"cat {amd_path} 2>/dev/null", default="")
+    if result == "1":
+        return "enabled"
+    elif result == "0":
+        return "disabled"
+
+    # ARM (Graviton) — no turbo mechanism
+    if platform.machine() in ("aarch64", "arm64"):
+        return "not_available"
+
+    return "unknown"
+
+
+def get_cpu_frequency_mhz() -> Optional[int]:
+    """Return current CPU frequency in MHz, or None if unavailable."""
+    freq_str = _run_cmd(
+        "cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_cur_freq 2>/dev/null",
+        default="",
+    )
+    if freq_str.isdigit():
+        return int(freq_str) // 1000  # kHz -> MHz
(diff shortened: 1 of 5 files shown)


process_metrics.py
```diff
@@ -34,6 +36,7 @@ def __init__(
         benchmark_threads: Optional[int] = None,
         architecture: Optional[str] = None,
         repository: Optional[str] = None,
+        environment_metadata: Optional[Dict[str, Any]] = None,
```
roshkhatri, line 40: I think we cannot have these as nested dict as we are converting them to columns , you would have to flatten the dict

valkey_benchmark.py
```diff
@@ -787,6 +788,11 @@ def _setup_profiling_and_metrics(self, profiling_set: dict, commit_time: str):
 
         metrics_processor = None
         if not profiling_enabled:
+            env_metadata = collect_environment_metadata(
+                benchmark_path=self.valkey_benchmark_path,
+                server_cpu_range=self.config.get("server_cpu_range"),
+                client_cpu_range=self.cores,
+            )
```
roshkhatri, line 795: stabilized is dead code in `collect_environment_metadata` no one calls it, so it will always stored as false

environment_metadata.py
```diff
@@ -0,0 +1,198 @@
...
+
+    if server_pid:
+        affinity = _run_cmd(
+            f"taskset -cp {server_pid} 2>/dev/null | grep -oP '(?<=: ).*'",
+            default="",
+        )
+        if affinity:
+            info["server_affinity"] = affinity
+
+    return info
```
roshkhatri: dead code? defined, but not used or tested

environment_metadata.py
```diff
@@ -0,0 +1,198 @@
...
+        "idle_states": get_idle_states_status(),
+        "aslr": get_aslr_status(),
+        "thp": get_thp_status(),
+        "numa_nodes": _run_cmd("lscpu | grep 'NUMA node(s)' | awk '{print $NF}'"),
+        "stabilized": stabilized,
+    }
+
+    freq = get_cpu_frequency_mhz()
+    if freq:
+        metadata["cpu_freq_mhz"] = freq
```
roshkhatri:
```suggestion
        metadata["cpu_freq_mhz_at_setup"] = freq
```

tests/test_environment_metadata.py
```diff
@@ -0,0 +1,152 @@
...
+            # Use real Path for the states_dir check
+            mock_path_cls.return_value = states_dir
+            # Can't easily mock Path glob, test the real function instead
+        # Just verify it returns a valid value
+        assert get_idle_states_status() in (
+            "all_disabled",
+            "partially_disabled",
+            "all_enabled",
+            "not_available",
+            "unknown",
```
roshkhatri: I dont understand this sorry, what are we doing here? I think the assertion cannot fail here?

environment_metadata.py
```diff
@@ -0,0 +1,198 @@
...
+        )
+        return result.stdout.strip() if result.returncode == 0 else default
+    except Exception:
+        return default
+
+
+def get_cpu_governor() -> str:
+    """Return the active CPU frequency governor, or 'not_available' for fixed-frequency CPUs."""
+    result = _run_cmd(
+        "cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_governor 2>/dev/null",
```
roshkhatri: Can we not Path("...").read_text()?

This run cmd with shell==True seems overkill with run command launching a shell to read a line?

process_metrics.py
rainsupreme: Flattened into env_-prefixed top-level keys in 1eb9bcc

valkey_benchmark.py
rainsupreme: Removed — will wire up properly in a future stabilization PR

environment_metadata.py
rainsupreme: good point!

tests/test_environment_metadata.py
rainsupreme: Replaced with mocked-sysfs tests that assert specific outcomes (turbo state, idle-states)

rainsupreme:
@roshkhatri I dropped the ball and forgot to reply/resolve the comments. 😅 anyway - ready for you to take another look!

environment_metadata.py
```diff
@@ -0,0 +1,190 @@
...
+    """Get the version string of the valkey-benchmark binary.
+
+    Runs `valkey-benchmark --version` and extracts the version + git SHA.
+    Falls back to 'unknown' if the binary doesn't exist or has no version output.
+    """
+    binary = Path(benchmark_path)
+    if not binary.exists():
+        return "unknown"
+
+    version_output = _run_cmd(f"{benchmark_path} --version", default="")
```
roshkhatri, line 124: `benchmark_path` is a user entered string, it can contain spaces  or special character, this would return `unknown` if that happens.
We would need to remove the `shell=true` for this.

The changes of this happening are quite thin, but just raising it

roshkhatri (approved):

PR #56, opened 2026-05-29
Add group, scenario, and description columns to metrics.json

Carol-ggg:
## Summary
Adds `group`, `scenario`, `scenario_type`, `group_description`, and `scenario_description` to `metrics.json` for `test_groups`-format runs. Failure markers also carry `group` and `scenario` for consistent grouping with success rows. Description fields are only emitted when configured. The simple/core config format is unaffected.

`utils/push_to_postgres.py` pins description columns to `VARCHAR(500)` so they don't bucket inconsistently from a single sample.

## Test plan
- [x] `black --check .` clean
- [x] `python -m pytest tests/` — 415 passed (404 existing + 11 new)
- [x] Captured `metrics.json` before/after running the same module test (`benchmark.py --module search ... --groups 1`); diff shows the 5 new fields populated on every row, no other changes
- [x] `TestIterateSimpleScenarios` confirms the new fields don't leak into runs that don't use `test_groups`

### DIFF
--- a/utils/push_to_postgres.py
+++ b/utils/push_to_postgres.py
@@ -80,6 +80,8 @@ def analyze_metrics_schema(metrics_data: List[Dict[str, Any]]) -> Dict[str, str]
             schema[field] = "TIMESTAMPTZ NOT NULL"
         elif field in ["commit", "command"]:
             schema[field] = f"VARCHAR(255) NOT NULL"
+        elif field in ["group_description", "scenario_description"]:
+            schema[field] = "VARCHAR(500)"
         else:
             sample_value = field_samples.get(field)
             column_type = detect_field_type(sample_value)
--- a/utils/push_to_postgres.py
+++ b/utils/push_to_postgres.py
@@ -259,6 +259,11 @@ def convert_metrics_to_rows(
                         row.append(None)
                 else:
                     row.append(None)
+            elif column in ["group_description", "scenario_description"]:
+                value = metric.get(column)
+                if isinstance(value, str) and len(value) > 500:
+                    value = value[:500]
+                row.append(value)
             else:
                 # Direct field mapping since field names are now normalized
                 row.append(metric.get(column))
--- a/utils/push_to_postgres.py
+++ b/utils/push_to_postgres.py
@@ -19,6 +19,8 @@ import psycopg2
 from psycopg2 import sql
 from psycopg2.extras import execute_values
 
+DESCRIPTION_MAX_LENGTH = 500
+
 
 def detect_field_type(value: Any) -> str:
     """Detect PostgreSQL column type from a sample value."""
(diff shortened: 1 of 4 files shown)


utils/push_to_postgres.py
```diff
@@ -80,6 +80,8 @@ def analyze_metrics_schema(metrics_data: List[Dict[str, Any]]) -> Dict[str, str]
             schema[field] = "TIMESTAMPTZ NOT NULL"
         elif field in ["commit", "command"]:
             schema[field] = f"VARCHAR(255) NOT NULL"
+        elif field in ["group_description", "scenario_description"]:
+            schema[field] = "VARCHAR(500)"
         else:
```
VoletiRam, line 87: Don't we need to add scenario and group id too?

Carol-ggg: Here we hardcoded the field type for group_description and scenario_description. 
For group, scenario, scenario type, it would be in the metrics dict through valkey_benchmark (_run_single_scenario), and the current logic would automatically detect the types for fields in the metrics.

VoletiRam: Makes sense. Thank you. Approving the PR.

VoletiRam (approved):

valkey_benchmark.py
```diff
@@ -945,6 +949,13 @@ def _run_single_scenario(
                     metrics["status"] = "success"
                     metrics["test_id"] = f"{group_id}_{scenario_id}"
                     metrics["test_phase"] = scenario_type
+                    metrics["group"] = group_id
+                    metrics["scenario"] = scenario_id
+                    metrics["scenario_type"] = scenario_type
```
roshkhatri: We have duplicate json fields here: `metrics["test_phase"]` and `metrics["scenario_type"]`
Should we consider dropping one or documenting why both exist?

valkey_benchmark.py
```diff
@@ -765,6 +766,8 @@ def _create_failure_marker(
         return {
             "test_id": f"{group_id}_{scenario_id}",
             "test_phase": scenario_type,
+            "group": group_id,
+            "scenario": scenario_id,
             "status": "failed",
             "error": error,
             "command": command,
```
roshkhatri, line 773: We are missing scenario_type, group_description, or scenario_description for failures, is it by design?

utils/push_to_postgres.py
roshkhatri: Why are we adding a cap of 500 char?
if we are adding this, how do we enforce it?

roshkhatri (commented):
Minor comments, LMK WDYT.

valkey_benchmark.py
Carol-ggg: this is by design, because we don't need additional description info for failure cases, we just need id to identify the test

utils/push_to_postgres.py
Carol-ggg: added truncation -  if length is longer than 500, only take first 500 char. 
added additional tests for this

valkey_benchmark.py
Carol-ggg: we deleted the scenario_type field

utils/push_to_postgres.py
```diff
@@ -80,6 +80,8 @@ def analyze_metrics_schema(metrics_data: List[Dict[str, Any]]) -> Dict[str, str]
             schema[field] = "TIMESTAMPTZ NOT NULL"
         elif field in ["commit", "command"]:
             schema[field] = f"VARCHAR(255) NOT NULL"
+        elif field in ["group_description", "scenario_description"]:
+            schema[field] = "VARCHAR(500)"
```
roshkhatri: can we make the 500 as a shared constant?

roshkhatri (commented):
Can you also resolve the DCO?\
everything looks good to merge

roshkhatri (approved):
LGTM

PR #57, opened 2026-06-04
 Add automated performance regression detection and notification

### LINKED ISSUE #30: Create an issue on valkey for a performance regression
When we see a performance regression, we should be automatically be able to open an issue on valkey

hanxizh9910:
## Summary
  
  Adds automated performance regression detection to the continuous benchmark workflow. Closes #30.
  
  After metrics are pushed to PostgreSQL, a new step compares the two most recent commits (by git commit time) using the existing CI-overlap method from `compare_benchmark_results.py`. If any configuration shows a statistically significant
  RPS regression above the threshold, a GitHub issue is opened on `valkey-io/valkey`.
  
  ## Changes
  
  - **`utils/detect_regression.py`**: queries DB for latest two commits, compares RPS metrics, outputs JSON, exits 1 if regression found
  - **`.github/workflows/valkey_benchmark.yml`**: added detection and issue creation steps after "Mark commits as complete"
  - **`tests/test_detect_regression.py`**: unit tests for the detection logic
  
  ## Configuration needed
  
  - `VALKEYRIE_BOT_APP_ID` / `VALKEYRIE_BOT_PRIVATE_KEY` — org secrets for cross-repo issue creation (same as valkey-ci-agent)
  - `REGRESSION_THRESHOLD_PCT` — repo variable (optional, default 5%)
  
  ## Deduplication
  
  Issues are deduplicated by title (`[PERF-REGRESSION] <new_sha> vs <baseline_sha>`). Before creating, the workflow checks for an existing open issue with the same title and skips if found.
  
  ## Testing
  
  - Unit tests pass (6 tests covering detection logic, DB conversion, threshold filtering)
  - Detection logic verified locally with Docker Postgres (dummy data, 15% drop → correctly detected)
  - Cross-repo issue creation verified end-to-end using PAT on fork ([example](https://github.com/hanxizh9910/valkey/issues/204))
  - Duplicate issue check verified (skips if issue already exists)
  - Tests re-run after latest changes, all passing


sarthakaggarwal97 (commented):
Some initial points and thoughts:
1. Can we add the links for commits or PR in the issue created, so the user can reach the commit easily?
2. In the sample [issue](https://github.com/hanxizh9910/valkey/issues/204) attached, it's mentioned `-15.0±0.3%`, we'd have exact numbers right, how much degradation we saw and in which case?
3. Should we attach some label to the issue? Maybe `performance`?
4. If other modules were to use the similar performance benchmark, which I think we are considering (@roshkhatri is the best person to know this), is the code reusable? Or should we consider moving this code to valkey-ci-agent?

sarthakaggarwal97:
Also, we can skip slack notifications for now. I think create an issue in valkey is enough.

.github/workflows/valkey_benchmark.yml
```diff
@@ -246,6 +247,136 @@ jobs:
             --config-file "$CONFIG_FILE" \
             $COMMIT_IDS
 
+      - name: Detect performance regression
+        id: detect_regression
+        continue-on-error: true
```
roshkhatri: Why continue on error?

.github/workflows/valkey_benchmark.yml
```diff
@@ -246,6 +247,136 @@ jobs:
...
+            exit 0
+          fi
+          report=$(echo "$report" | python -c "import sys,json; print(json.dumps(json.load(sys.stdin)))")
+          echo "report=$report" >> "$GITHUB_OUTPUT"
+          echo "has_regression=$(echo "$report" | python -c "import sys,json; print(str(json.load(sys.stdin)['has_regression']).lower())")" >> "$GITHUB_OUTPUT"
+          echo "new_commit=$(echo "$report" | python -c "import sys,json; print(json.load(sys.stdin)['new_commit'][:8])")" >> "$GITHUB_OUTPUT"
+          echo "baseline_commit=$(echo "$report" | python -c "import sys,json; print(json.load(sys.stdin)['baseline_commit'][:8])")" >> "$GITHUB_OUTPUT"
+
+      - name: Generate bot token
+        id: bot-token
```
roshkhatri, line 283:
```suggestion
        id: bot-token
        if: steps.detect_regression.outputs.has_regression == 'true'
```

roshkhatri: Also if repo owner is valkey-io

utils/detect_regression.py
```diff
@@ -0,0 +1,167 @@
...
+def main() -> None:
+    parser = argparse.ArgumentParser(
+        description="Detect performance regressions from PostgreSQL benchmark data"
+    )
+    parser.add_argument("--host", required=True)
+    parser.add_argument("--port", type=int, default=5432)
+    parser.add_argument("--database", required=True)
+    parser.add_argument("--username", required=True)
+    parser.add_argument("--password", required=True)
+    parser.add_argument("--table-name", default="benchmark_metrics")
```
roshkhatri: table name should be required, and not defaulting to one

roshkhatri (commented):
good start, need to add unit tests for detect_regression.

roshkhatri:
@sarthakaggarwal97 
> Should we attach some label to the issue? Maybe performance?

It is already added in the PR

> If other modules were to use the similar performance benchmark, which I think we are considering (@roshkhatri is the best person to know this), is the code reusable? Or should we consider moving this code to valkey-ci-agent?

the Modules and anyone can use the same benchmark workflow to add this detection, moving this code to `valkey-ci-agent` we will have to query the db the get all the datapoints and then see where we are seeing a regression.

I think Modules will own their workflows for benchmarking, but they can pick this code from here.

hanxizh9910:
> @sarthakaggarwal97
> 
> > Should we attach some label to the issue? Maybe performance?
> 
> It is already added in the PR
> 
> > If other modules were to use the similar performance benchmark, which I think we are considering (@roshkhatri is the best person to know this), is the code reusable? Or should we consider moving this code to valkey-ci-agent?
> 
> the Modules and anyone can use the same benchmark workflow to add this detection, moving this code to `valkey-ci-agent` we will have to query the db the get all the datapoints and then see where we are seeing a regression.
> 
> I think Modules will own their workflows for benchmarking, but they can pick this code from here.

I think the intern Carol Guo is working on a benchmark project like Valkey-perf-benchmark for the Modules team

.github/workflows/valkey_benchmark.yml
hanxizh9910: Removed!

.github/workflows/valkey_benchmark.yml
```diff
@@ -246,6 +247,125 @@ jobs:
...
+        env:
+          REGRESSION_THRESHOLD: ${{ vars.REGRESSION_THRESHOLD_PCT || '5' }}
+        run: |
+          PGPASSWORD=$(aws rds generate-db-auth-token \
+            --hostname ${{ secrets.POSTGRESQL_ENDPOINT }} \
+            --port 5432 \
+            --region ${{ secrets.AWS_REGION }} \
+            --username ${{ secrets.RDS_USERNAME }})
+          set +e
+          report=$(python utils/detect_regression.py \
```
roshkhatri: Should we change this to python -m utils.detect_regression ? the wf that tested your change in your repo uses this. have you tested the workflow again?

.github/workflows/valkey_benchmark.yml
```diff
@@ -246,6 +247,125 @@ jobs:
...
+        if: steps.detect_regression.outputs.has_regression == 'true' && github.repository_owner == 'valkey-io'
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          client-id: ${{ secrets.VALKEYRIE_BOT_APP_ID }}
+          private-key: ${{ secrets.VALKEYRIE_BOT_PRIVATE_KEY }}
+          owner: valkey-io
+          repositories: valkey
+
+      - name: Open GitHub issue for regression
+        if: steps.detect_regression.outputs.has_regression == 'true'
```
roshkhatri: should also ad the owner guard here

utils/detect_regression.py
```diff
@@ -0,0 +1,167 @@
...
+
+    baseline_data = average_multiple_runs(
+        fetch_metrics_for_commit(conn, table_name, baseline_sha)
+    )
+    new_data = average_multiple_runs(
+        fetch_metrics_for_commit(conn, table_name, new_sha)
+    )
+
+    config_groups, baseline_version, new_version, _, _ = create_comparison_table_data(
+        baseline_data, new_data
```
roshkhatri: we need to add a filter for `rps` here

utils/detect_regression.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def detect(
+    conn: psycopg2.extensions.connection,
+    table_name: str,
+    threshold_pct: float,
+    test_type: str = "core",
+) -> Dict[str, Any]:
+    """Compare the two most recent commits and return a regression report dict."""
+    new_sha, baseline_sha = fetch_last_two_commits(conn, table_name, test_type)
```
roshkhatri, line 87: This this taking the commit time or the benchmark time?

.github/workflows/valkey_benchmark.yml
```diff
@@ -35,6 +35,7 @@ defaults:
 permissions:
   id-token: write
   contents: read
+  issues: write
```
roshkhatri: This should be unused as we are using the bot token

roshkhatri (commented):
How are we going to de-dup the issues for a specific commit-id?
Can you also please run the tests again and update the top comment?

utils/detect_regression.py
hanxizh9910: It's the commit time: https://github.com/valkey-io/valkey-perf-benchmark/blob/main/process_metrics.py#L93

hanxizh9910:
> How are we going to de-dup the issues for a specific commit-id? Can you also please run the tests again and update the top comment?

Hi, I re-run tests after latest changes, all passing. Issue creation works ([example](https://github.com/hanxizh9910/valkey/issues/208)) and deduplication confirmed [second run skips with 'Issue already exists', line 85 of step `Open Github issue for Regression`](https://github.com/hanxizh9910/valkey-perf-benchmark/actions/runs/27852625694/job/82434313210). Tested with PAT on my fork since Valkeyrie Bot secrets are org-level only

.github/workflows/valkey_benchmark.yml
```diff
@@ -246,6 +246,125 @@ jobs:
...
+              if (e.status === 404) {
+                await github.rest.issues.createLabel({
+                  owner: targetOwner,
+                  repo: targetRepo,
+                  name: label,
+                  color: 'e11d48',
+                  description: 'Performance regression detected by CI',
+                });
+              }
+            }
```
sarthakaggarwal97: I think the app doesn't have the permission to create labels. This is going to be dead code. We can just create label manually once.

sarthakaggarwal97 (approved):
This looks good to me. One minor comment. Thanks @hanxizh9910 for the change and thanks @roshkhatri for reviewing it.

.github/workflows/valkey_benchmark.yml
hanxizh9910: Done, just removed the label creation block. We'll create the performance-regression label manually

roshkhatri (approved):
I think all the comments are addressed, Good to merge, thanks @sarthakaggarwal97 for reviewing it again

PR #58, opened 2026-06-12
Add module benchmark commit tracking and Grafana dashboard

Carol-ggg:
## Summary

- Add `module_postgres_track_commits.py` — a queue-based 2D commit tracking system that tracks (core_sha, module_sha) pairs across both valkey and module repositories, with config_set-aware subset detection and priority classification
- Extend `push_to_postgres.py` to support module benchmarks: auto-generate table names from `--module` flag, store `config_set` as JSONB, support `--max-timestamp` for 2D commit ordering
- Add Grafana dashboard JSON for module benchmarks with template variables (module, config|arch, config_set, group, scenario), commit annotation markers, and data links to GitHub commits
- Add comprehensive unit tests and integration tests for all new functionality

## Changes

**New files:**
- `utils/module_postgres_track_commits.py` — Queue-based module commit tracking with populate, fetch-next, mark-complete, cleanup, and check-incomplete operations. Supports config_sets as JSONB, subset detection (marks pairs as completed_as_subset when a superset config is ran), and priority assignment (forward vs fallback based on newest completed pair)
- `dashboards/grafana/module/module-dashboard_1.json` — Grafana dashboard with 6 time-series panels (RPS, p99, p95, p50, min, max latency), debug tables, cascading template variables, and commit data links
- `tests/test_module_postgres_track_commits.py` — Unit tests for pure logic (config_sets subset detection, table naming, CommitPair validation)
- `tests/integration/test_module_tracking_integration.py` — Integration tests against real PostgreSQL covering the full lifecycle and subset detection scenarios

**Modified files:**
- `utils/push_to_postgres.py` — Table name auto-generation from `--module`, JSONB support for `config_set`, `--max-timestamp` override, `resolve_table_name()` helper
- `utils/postgres_track_commits.py` — Minor adjustments for module tracking compatibility
- `benchmark.py` / `valkey_benchmark.py` — Pass config_name through to metrics output
- `tests/test_postgres_utils.py` — Tests for new schema detection (JSONB, config_name, resolve_table_name), also updated test on `description_max_length` from last PR

## Test plan

- [ ] Unit tests pass: `pytest tests/test_module_postgres_track_commits.py tests/test_postgres_utils.py -v`
- [ ] Integration tests pass (requires Docker Postgres on port 5433): `pytest tests/integration/test_module_tracking_integration.py -v`
- [ ] End-to-end: trigger workflow from valkey-search repo, verify metrics land in Postgres and Grafana dashboard renders with data

## Dashboard
- **Filter dropdowns** — Select module, config/architecture, config set, group, scenario, and cluster mode to slice the data.
- **Performance graphs** — RPS and latency (p50/p95/p99/min/max) over time, with clickable commit links to GitHub.
- **Regression table** — Auto-flags runs where RPS dropped or p99 spiked compared to the previous run.

Screenshot 2026-07-01 at 11 30 49 PM
Screenshot 2026-07-01 at 11 31 35 PM
Screenshot 2026-07-01 at 11 31 44 PM
Screenshot 2026-07-01 at 11 31 49 PM


tests/integration/test_benchmark_execution.py
```diff
@@ -14,7 +14,6 @@
 from valkey_benchmark import ClientRunner
 from process_metrics import MetricsProcessor
 
-
```
VoletiRam, line 17: Lets avoid unecessary touch points

tests/integration/test_git_ref_resolution.py
```diff
@@ -12,7 +12,6 @@
 
 from utils.git_utils import resolve_ref, get_commit_timestamp
 
-
```
VoletiRam, line 15: Lets avoid unecessary touch points

utils/module_postgres_track_commits.py
```diff
@@ -0,0 +1,981 @@
...
+            (e.g., '../valkey-search/.github/benchmark_configs/fts-benchmarks-arm.json')
+
+    Returns:
+        Just the filename (e.g., 'fts-benchmarks-arm.json')
+    """
+    return Path(config_file_path).name
+
+
+def _module_table_name(module_name: str) -> str:
+    """Return the tracking table name for a given module."""
```
VoletiRam, line 147: Can we validate the input for module name here?  There is a chance that If someone passes `search; DROP TABLE benchmark_commits;`, we have risk of losing data.


__Fix:__ Validate against a strict regex:

```python
import re
def _module_table_name(module_name: str) -> str:
    if not re.match(r'^[a-z][a-z0-9_]{0,30}$', module_name):
        raise ValueError(f"Invalid module_name: {module_name}")
    return f"benchmark_module_commits_{module_name}"
```

utils/module_postgres_track_commits.py
```diff
@@ -0,0 +1,981 @@
...
+            "architecture": self.architecture,
+        }
+        missing = [k for k, v in required.items() if v is None or v == ""]
+        if missing:
+            print(
+                f"FATAL: CommitPair missing required fields: {missing} "
+                f"(pair={self.core_sha}:{self.module_sha})",
+                file=sys.stderr,
+            )
+            sys.exit(1)
```
VoletiRam: Instead of exiting on validation failure in dataclass, can we raise Raise `ValueError`? Let `main()` catch it:

```python
def __post_init__(self):
    missing = [k for k, v in required.items() if v is None or v == ""]
    if missing:
        raise ValueError(f"CommitPair missing required fields: 
```



The reason is It makes the class impossible to use in any context other than a CLI script. The tests even have to use `pytest.raises(SystemExit)` which proves the issue.

utils/module_postgres_track_commits.py
```diff
@@ -0,0 +1,981 @@
...
+from dataclasses import dataclass
+from pathlib import Path
+from typing import List, Optional
+
+import psycopg2
+from psycopg2.extras import Json
+
+from datetime import datetime
+
+from postgres_track_commits import _git_rev_list, _git_commit_time
```
VoletiRam: These are private methods in postgres_track_commits being imported across modules. This breaks encapsulation and creates a hidden coupling that could break if `postgres_track_commits.py` is refactored.

I notice we already have utils `utils/git_utils.py`. Move `_git_rev_list` and `_git_commit_time` there as public functions, import from that shared module in both files.

utils/module_postgres_track_commits.py
```diff
@@ -0,0 +1,981 @@
...
+                pair.priority = 2
+                fallback_count += 1
+
+    print(
+        f"Priority assigned in memory: {forward_count} forward, {fallback_count} fallback",
+        file=sys.stderr,
+    )
+
+
+def populate_module_commits(
```
VoletiRam, line 392: This single function creates table, Runs git rev-list on two repos, Fetches all timestamps (subprocess per commit!), Queries existing pairs from DB, Builds CommitPair objects (cartesian product), Runs subset detection, Assigns priority, and Batch inserts. If you have 50 core commits × 20 module commits, that's 70 subprocess calls just for timestamps. With large repos, this is a performance landmine.

At minimum, the timestamp fetching should use `git log --format="%H %cI"` to get all timestamps in a single subprocess call per repo.

utils/module_postgres_track_commits.py
```diff
@@ -0,0 +1,981 @@
...
+    """
+    if not isinstance(subset, list) or not isinstance(superset, list):
+        return False
+    for item in subset:
+        if item not in superset:
+            return False
+    return True
+
+
+def _create_module_table(conn, module_name: str) -> None:
```
VoletiRam, line 176: This is getting called in every function. Every function (`fetch_next_module_commits`, `mark_module_commits`, `cleanup_module_commits`, `check_incomplete_rows`) starts with `_create_module_table(conn, module_name)`. This means every single CLI call issues a `CREATE TABLE IF NOT EXISTS` + 7 `CREATE INDEX IF NOT EXISTS` statements.

Can we Call it once in `main()` after connection, before dispatching to the operation?

utils/module_postgres_track_commits.py
```diff
@@ -0,0 +1,981 @@
...
+        for superset_cs in superset_list:
+            cur.execute(
+                f"""
+                SELECT sha, module_sha FROM {table}
+                WHERE status = 'completed'
+                  AND config_name = %s AND architecture = %s
+                  AND config_sets = %s
+            """,
+                (config_name, architecture, Json(superset_cs)),
+            )
```
VoletiRam: Hmm, here we are doing N Queries Per Superset Config. Though we currently have two sets, it might increase increase in future. Can we use a single query with `config_sets = ANY(...)` or a `WHERE config_sets IN (...)` instead?

utils/push_to_postgres.py
```diff
@@ -386,13 +403,13 @@ def process_commit_metrics(
...
-    # Augment metrics with test_type, module, and module_commit (extension for FTS tests)
+    # Augment metrics with test_type, module_commit, and timestamp override at push time
     for metric in metrics_data:
         metric["test_type"] = test_type
-        if module:
-            metric["module"] = module
         if module_commit:
             metric["module_commit"] = module_commit
+        if max_timestamp:
+            metric["timestamp"] = max_timestamp
```
VoletiRam, line 411: Why are we overriding the timestamp here? If my understanding is right, here we are overriding the commit timestamp with computed value.

utils/module_postgres_track_commits.py
```diff
@@ -0,0 +1,981 @@
...
+    if not_ready:
+        print(
+            f"FATAL: {len(not_ready)} pairs not ready to insert "
+            f"(missing priority or status). Aborting.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
+    # Step 5: Batch insert — single transaction, all or nothing
+    from psycopg2.extras import execute_values
```
VoletiRam: Can we move imports to the top of the file?

utils/module_postgres_track_commits.py
```diff
@@ -0,0 +1,981 @@
...
+
+    insert_sql = f"""
+        INSERT INTO {table} (sha, module_sha, core_timestamp,
+                             module_timestamp, max_commit_timestamp,
+                             min_commit_timestamp, status, priority,
+                             config_name, config_sets, architecture)
+        VALUES %s
+    """
```
VoletiRam, line 525: Missing Idempotent. if two workflows race and try to insert at the same time, this will crash with `IntegrityError` and exit. The code claims "existing pairs were filtered beforehand" but there's a TOCTOU race between the `SELECT DISTINCT` and the `INSERT`.


Add `ON CONFLICT (sha, module_sha, config_name, config_sets, architecture) DO NOTHING` to make it safe.

utils/push_to_postgres.py
```diff
@@ -266,6 +273,16 @@ def convert_metrics_to_rows(
                 if isinstance(value, str) and len(value) > DESCRIPTION_MAX_LENGTH:
                     value = value[:DESCRIPTION_MAX_LENGTH]
                 row.append(value)
+            elif column == "config_set":
+                from psycopg2.extras import Json
```
VoletiRam: Move imports to top. Also, this is already imported i guess.

tests/test_module_postgres_track_commits.py
```diff
@@ -0,0 +1,243 @@
...
+# ---------------------------------------------------------------------------
+
+
+class TestCommitPair:
+    """Tests for CommitPair validation and methods."""
+
+    def _make_pair(self, **overrides):
+        """Helper to create a valid CommitPair with optional overrides."""
+        from datetime import datetime, timezone
+        from utils.module_postgres_track_commits import _parse_timestamp
```
VoletiRam: Move imports to top of the file. Please do it for other places where we added them in middle.

utils/module_postgres_track_commits.py
```diff
@@ -0,0 +1,981 @@
...
+
+    with conn.cursor() as cur:
+        cur.execute(
+            f"""
+            SELECT COUNT(*) FROM {table}
+            WHERE config_name = %s AND config_sets = %s AND architecture = %s
+              AND (sha IS NULL OR module_sha IS NULL OR core_timestamp IS NULL
+                   OR module_timestamp IS NULL OR max_commit_timestamp IS NULL
+                   OR min_commit_timestamp IS NULL OR status IS NULL
+                   OR priority IS NULL)
```
VoletiRam: here we are checking for NULL but the table schema has `sha VARCHAR(40) NOT NULL`, `module_sha VARCHAR(40) NOT NULL`, etc. Those NULLs can never exist by database constraint. The only field that could be NULL is `priority INTEGER` (no NOT NULL constraint). Do we need this check(and function)?

utils/push_to_postgres.py
```diff
@@ -386,13 +403,13 @@ def process_commit_metrics(
         print(f"Skipping {commit_dir.name}: empty metrics")
         return 0, True
 
-    # Augment metrics with test_type, module, and module_commit (extension for FTS tests)
+    # Augment metrics with test_type, module_commit, and timestamp override at push time
     for metric in metrics_data:
         metric["test_type"] = test_type
-        if module:
```
VoletiRam, line 392: Why are we dropping this? Don't we lose the option to select module from grafana dashboard if we do this?

utils/module_postgres_track_commits.py
```diff
@@ -0,0 +1,981 @@
...
+    conn,
+    repo: Path,
+    branch: str,
+    module_repo: Path,
+    module_branch: str,
+    architecture: str,
+    module_name: str,
+    config_name: str,
+    config_sets: List[dict],
+    config_sets_json,
```
VoletiRam: Instead of passing both raw and wrapped version of config_set to this method, can we pass only `config_sets` (the raw list), and wrap it at SQL query time inside each function that needs it. One source of truth.

tests/integration/test_benchmark_execution.py
Carol-ggg: I reverted the change.

tests/integration/test_git_ref_resolution.py
Carol-ggg: reverted the change

Carol-ggg: resolved in commit 84140602ffb6e6f60df992204a121679924267b8

tests/integration/test_benchmark_execution.py
Carol-ggg: resolved in 2ecab1b16705431d5def2318139674b9af84a54f

utils/module_postgres_track_commits.py
Carol-ggg: thanks so much! that's definitely a good catch. i added this logic and also added related tests. changes in  commit 654bc85f164c75cdfa5a6049888e16233d02bd84

tests/test_module_postgres_track_commits.py
```diff
@@ -0,0 +1,255 @@
...
+        assert _module_table_name("search") == "benchmark_module_commits_search"
+
+    def test_json_module(self):
+        assert _module_table_name("json") == "benchmark_module_commits_json"
+
+    def test_bloom_module(self):
+        assert _module_table_name("bloom") == "benchmark_module_commits_bloom"
+
+    def test_arbitrary_name(self):
+        assert _module_table_name("my_module") == "benchmark_module_commits_my_module"
```
VoletiRam: These are testing same piece of code. Do you see value in testing this logic which seems to be just prefixing the string? I prefer removing this since it is redundant and cover one case in other high level test case instead.

utils/module_postgres_track_commits.py
```diff
@@ -0,0 +1,977 @@
...
+    """Extract the config file name from a full path.
+
+    Args:
+        config_file_path: Full or relative path to config file
+            (e.g., '../valkey-search/.github/benchmark_configs/fts-benchmarks-arm.json')
+
+    Returns:
+        Just the filename (e.g., 'fts-benchmarks-arm.json')
+    """
+    return Path(config_file_path).name
```
VoletiRam, line 143: Testing the library is not needed. We should focus on testing what we added and avoid necessary coverage.

tests/test_module_postgres_track_commits.py
```diff
@@ -0,0 +1,255 @@
...
+# config_set JSONB in module commit tracking
+# ---------------------------------------------------------------------------
+
+
+class TestModuleConfigSetsJsonb:
+    """Test that config_sets array is wrapped as Json for JSONB insertion."""
+
+    def test_config_sets_array_wrapped_as_json(self):
+        from psycopg2.extras import Json
+
```
VoletiRam: We don't need to test library logic, only the logic we are adding explicitly. we are validating if config set can be converted to json or not and validating the library functionality here which is unecessary. We can remove all the tests that is testing the library logic (likely whole tests in this class)

tests/test_postgres_utils.py
```diff
@@ -340,5 +347,134 @@ def test_descriptions_over_500_chars_truncated(self):
...
+
+
+class TestConfigSetJsonb:
+    """Tests for config_set being stored as JSONB."""
+
+    def test_config_set_stored_as_jsonb(self):
+        from psycopg2.extras import Json
+
+        metrics = [
+            {
```
VoletiRam: Same reasoning as other. Testing library functionality. Please remove these tests.

tests/integration/test_module_tracking_integration.py
```diff
@@ -0,0 +1,2281 @@
...
+
+# ---------------------------------------------------------------------------
+# Timestamp computation
+# ---------------------------------------------------------------------------
+
+
+class TestTimestampComputation:
+    """Verify max_commit_timestamp and min_commit_timestamp are computed correctly."""
+
+    def test_max_min_timestamps_computed(self, conn, mock_git):
```
VoletiRam: isn't this test duplicate the `test_inserts_cartesian_product`? The other tests alreadu covers the min and max timestamp logic. We can remove this since redundant.

tests/integration/test_module_tracking_integration.py
```diff
@@ -0,0 +1,2281 @@
...
+            )
+            row = cur.fetchone()
+
+        core_ts, mod_ts, max_ts, min_ts = row
+        assert core_ts.isoformat() == "2026-06-05T10:00:00+00:00"
+        assert mod_ts.isoformat() == "2026-06-01T10:00:00+00:00"
+        assert max_ts.isoformat() == "2026-06-05T10:00:00+00:00"
+        assert min_ts.isoformat() == "2026-06-01T10:00:00+00:00"
+
+    def test_max_min_when_module_is_newer(self, conn, mock_git):
```
VoletiRam: Same reasoning. isn't this test duplicate the test_inserts_cartesian_product? The other tests already covers the min and max timestamp logic. We can remove this since redundant.

tests/integration/test_module_tracking_integration.py
```diff
@@ -0,0 +1,2281 @@
...
+            assert status == "pending"
+            assert config_name == CONFIG_NAME
+            assert architecture == ARCHITECTURE
+            assert priority == 1  # no pointer yet → all forward
+            assert core_ts.isoformat() == exp_core_ts
+            assert mod_ts.isoformat() == exp_mod_ts
+            assert max_ts.isoformat() == exp_max_ts
+            assert min_ts.isoformat() == exp_min_ts
+
+    def test_skips_existing_pairs(self, conn, mock_git):
```
VoletiRam: Isn't this test duplicate the `test_repopulate_large_set_inserts_zero` ?

Both test the same thing (idempotent re-insert), just at different scales (1×1 vs 10×10). The 10×10 version adds no new code path coverage. We can keep either 1

utils/module_postgres_track_commits.py
Carol-ggg: makes sense. I have changed this to `valueError` and I have changed other `sys.exit` usage in this file. I added in `main()` to catch the `valueerror` and exit. I also updated the test files accordingly. 
the changes are in commit d05089ea72dc8c756d6870b01a8b47a97eb7881e

tests/integration/test_module_tracking_integration.py
```diff
@@ -0,0 +1,2281 @@
...
+                "SELECT EXISTS (SELECT FROM information_schema.tables WHERE table_name = %s)",
+                (table,),
+            )
+            assert cur.fetchone()[0] is True
+
+    def test_idempotent(self, conn):
+        """Calling _create_module_table twice should not error."""
+        _create_module_table(conn, MODULE_NAME)
+        _create_module_table(conn, MODULE_NAME)
+
```
VoletiRam: Here we are testing postgres functionality of CREATE command. Lets remove them since they add less value.

tests/test_client_runner_logic.py
```diff
@@ -355,6 +355,44 @@ def test_marker_includes_group_and_scenario(self, minimal_client_runner):
         assert marker["scenario"] == "search_idx"
 
 
+# ---------------------------------------------------------------------------
+# config_name conditional output
+# ---------------------------------------------------------------------------
+
+
+class TestConfigName:
+    """Verify config_name is only present when provided (module runs only)."""
```
VoletiRam: These two tests are just testing set/get logic alone which has low value. Can be removed.

tests/integration/test_module_tracking_integration.py
VoletiRam, line 1: Many tests have multi-line docstring comments explaining what the test does. For simple tests, the test name IS the documentation. `test_fetches_pending_pairs` + the code is self-evident. Remove comments from obvious tests. Keep them only for complex scenarios (like the priority classification or subset detection tests).

VoletiRam, line 1: Both `clean_table` and `mock_git` are `autouse=True`. The `mock_git` auto-mocking means that even `TestCreateModuleTable.test_creates_table` (which doesn't need git mocks at all) gets the mock injected.  Tests should only get the fixtures they need. Can you try removing unwanted injections?

tests/test_module_postgres_track_commits.py
VoletiRam, line 1: Missing tests for `_parse_timestamp` edge cases.What happens with naive timestamps? Malformed strings? This function is used everywhere.

Verify `config_name` appears in JSON when set (not just the runner attribute)

VoletiRam (commented):
Summary of the tests that can be deleted:

Test | Reason
-- | --
3 of 4 TestModuleTableName happy-path | Same f-string
5 of 6 TestGetConfigName | Testing Path.name
TestModuleConfigSetsJsonb class entirely | Testing psycopg2
TestTimestampComputation (both tests) | Duplicate of test_inserts_cartesian_product
test_repopulate_large_set_inserts_zero | Duplicate of test_skips_existing_pairs
TestPriorityIntegration.test_first_run_all_forward | Duplicate of direct unit test
TestPopulateIntegration.test_all_rows_get_priority_assigned | Triple-covered
TestConfigSetJsonb psycopg2 assertions | Third-party testing

utils/module_postgres_track_commits.py
Carol-ggg: that makes sense. i have moved both functions to git_utils and import to `module_postgres_track_commits.py` and `postgres_track_commits.py` accordingly
changes reflected in 3d34e0de0f0dabe96ec9f5f80ae2ebf12799c85f

utils/module_postgres_track_commits.py
Carol-ggg: i have updated all imports to be at the top of the file. changes reflected in 3d34e0de0f0dabe96ec9f5f80ae2ebf12799c85f

utils/push_to_postgres.py
Carol-ggg: this is fixed in 3d34e0de0f0dabe96ec9f5f80ae2ebf12799c85f

tests/test_module_postgres_track_commits.py
Carol-ggg: this is fixed in 3d34e0de0f0dabe96ec9f5f80ae2ebf12799c85f

utils/module_postgres_track_commits.py
Carol-ggg: i added a function `git_rev_list_with_timestamps` in `git_utils.py` to grab time and commit together so it will be one subprocess per repo. The changes are reflected in 2d1539f8bce454cce5c06483dccb6bf2f064d7da
one thing worth noting is that, I have previously moved `giv_rev_list` and `git_commit_time` to `git_utils.py`. However, these two functions are only used in the `postgres_track_commits.py `file after this change. It could exist as a private helper function in that file. Do we want to revert them back to private helper functions? 
The `git_rev_list_with_timestamps`  could also be a private helper function in the `module_postgres_track_commit`.py file.

tests/test_module_postgres_track_commits.py
Carol-ggg: keep only `test_arbitrary_name` and removed the other three. changes in commit 2d1539f8bce454cce5c06483dccb6bf2f064d7da

utils/module_postgres_track_commits.py
Carol-ggg: deleted the 5/6 of the test config name tests and kept only 1. the changes is reflected in 2d1539f8bce454cce5c06483dccb6bf2f064d7da

tests/test_module_postgres_track_commits.py
Carol-ggg: this test is deleted, the changes are reflected in 2d1539f8bce454cce5c06483dccb6bf2f064d7da

tests/test_postgres_utils.py
Carol-ggg: deleted. changes in 2d1539f8bce454cce5c06483dccb6bf2f064d7da

tests/integration/test_module_tracking_integration.py
Carol-ggg: deleted this test. changes in 2d1539f8bce454cce5c06483dccb6bf2f064d7da

tests/integration/test_module_tracking_integration.py
Carol-ggg: deleted. changes in 2d1539f8bce454cce5c06483dccb6bf2f064d7da

tests/integration/test_module_tracking_integration.py
Carol-ggg: deleted this function, and kept the `test_repopulate_large_set_inserts_zero`function 
changes in 2d1539f8bce454cce5c06483dccb6bf2f064d7da

tests/integration/test_module_tracking_integration.py
Carol-ggg: This test is removed. changes reflected in 2d1539f8bce454cce5c06483dccb6bf2f064d7da

tests/test_client_runner_logic.py
Carol-ggg: removed. changes in 2d1539f8bce454cce5c06483dccb6bf2f064d7da

tests/integration/test_module_tracking_integration.py
Carol-ggg: I have removed unnecessary comments and make the comments overall more concise. changes in 2d1539f8bce454cce5c06483dccb6bf2f064d7da

Carol-ggg:
# Testcases Deleted

changes reflected in 2d1539f8bce454cce5c06483dccb6bf2f064d7da and 114a186bcd19d317c0e411259c466f375797c1c5

## test_module_postgres_track_commits.py (10 deleted)

| #   | Test Name                                           | Notes                                                  |
| -----| -----------------------------------------------------| --------------------------------------------------------|
| 1   | `test_extracts_filename_from_absolute_path`         | Keep only 1 of 6 `TestGetConfigName`                   |
| 2   | `test_extracts_filename_when_just_filename`         | Keep only 1 of 6 `TestGetConfigName`                   |
| 3   | `test_handles_deeply_nested_dirs`                   | Keep only 1 of 6 `TestGetConfigName`                   |
| 4   | `test_handles_dotfiles`                             | Keep only 1 of 6 `TestGetConfigName`                   |
| 5   | `test_preserves_extension`                          | Keep only 1 of 6 `TestGetConfigName`                   |
| 6   | `test_search_module`                                | Keep only 1 of 4 `TestModuleTableName`                 |
| 7   | `test_json_module`                                  | Keep only 1 of 4 `TestModuleTableName`                 |
| 8   | `test_bloom_module`                                 | Keep only 1 of 4 `TestModuleTableName`                 |
| 9   | `test_config_sets_array_wrapped_as_json`            | Deleted `TestModuleConfigSetsJsonb` - testing psycopg2 |
| 10  | `test_none_config_sets_defaults_to_empty_dict_list` | Deleted `TestModuleConfigSetsJsonb` - testing psycopg2 |
| 11  | `test_creates_table` | Tested implicitly in other tests |


## test_client_runner_logic.py (2 deleted)

| #   | Test Name                              | Notes                                 |
| -----| ----------------------------------------| ---------------------------------------|
| 12  | `test_config_name_present_when_set`    | Just testing set/get logic, low value |
| 13  | `test_config_name_absent_when_not_set` | Just testing set/get logic, low value |

## test_module_tracking_integration.py (5 deleted)

| #   | Test Name                             | Notes                                                                             |
| -----| ---------------------------------------| -----------------------------------------------------------------------------------|
| 14  | `test_idempotent`                     | Testing postgres functionality of CREATE command, less value                      |
| 15  | `test_skips_existing_pairs`           | Duplicate of `test_repopulate_large_set_inserts_zero`                             |
| 16  | `test_first_run_all_forward`          | Duplicate of `test_no_pointer_all_forward`                                        |
| 17  | `test_max_min_timestamps_computed`    | Deleted `TestTimestampComputation`, duplicate of `test_inserts_cartesian_product` |
| 18  | `test_max_min_when_module_is_newer`   | Deleted `TestTimestampComputation`, duplicate of `test_inserts_cartesian_product` |
| 19  | `test_all_rows_get_priority_assigned` | Duplicate of `test_no_pointer_all_forward`                                        |


## test_postgres_utils.py (1 deleted)

| #   | Test Name                         | Notes                                                  |
| -----| -----------------------------------| --------------------------------------------------------|
| 20  | `test_config_set_stored_as_jsonb` | Deleted `TestModuleConfigSetsJsonb` - testing psycopg2 |

tests/integration/test_module_tracking_integration.py
Carol-ggg: i removed the `autouse=True `for `mock_git` and only pass it in for functions that actually need it. For `clean_table`, I leave the` autouse=True` as create table is used in every function. I removed the `test_creates_table` function as this is implicitly tested in other tests. 
changes in 533ed95aad77c5c6fd2be8eac3af79358f266ec6

utils/module_postgres_track_commits.py
Carol-ggg: changed accordingly, only called in main once before dispatch. changes in 23f6b845bed504014cb500faf8e39c13d69b3261

utils/push_to_postgres.py
Carol-ggg: In our current design, each module has its own table. Including the module column would mean that the table would have the same word for the whole column.

utils/push_to_postgres.py
Carol-ggg: `timestamp` is mainly used for grafana queries. It is used as the x-axis of the visualization. 
In core's pipeline, they directly use the core commit timestamp as the `timestamp` and use that for grafana. 
However, for our testing, we would have both module and core commit timestamp. Using `max_timestamp` (deduced and passed in from our workflow) will give us the newer timestamp from core and module. I think this would make more sense for our visualization.

utils/module_postgres_track_commits.py
Carol-ggg: I have updated the query to use `IN(..)`. I have also updated the subset test cases so it tests when we have more than one superset to verify the new query works. changes in 709bef507424c345a9438eb2d1e2d30cb5f6777b

utils/module_postgres_track_commits.py
Carol-ggg: I have updated the functions. Now all the functions takes in config_sets and they would wrap it to json format if needed. changes in 709bef507424c345a9438eb2d1e2d30cb5f6777b

utils/module_postgres_track_commits.py
Carol-ggg: That's a good point. I can definitely add this as a fall back. 
For real production, I am wondering will we consider using two runners for the same config file and architecture? Depends on the answer to this, we could consider having just a log or raise some error.

utils/module_postgres_track_commits.py
Carol-ggg: this function is mainly defensive. You are right that it is kind of redundant, as we already check priority NOT NULL using `is_ready_to_insert` before inserting.  
I have deleted this function and updated accordingly. Also updated the workflow file in valkeysearch repo. 
The changes are in 94fdd78155c1f9f0e27eb65bd272cf90d436afa6

tests/test_module_postgres_track_commits.py
Carol-ggg: add time_stamp related testings. Also updated the timestamp logic so that now max_timestamp is passed from previous steps. The changes are reflected in 59ce9443c885da8328c88c7df5f191f2f9ca2510

Carol-ggg: It's hard to add a full end-to-end test for `config_name` at the runner level because it's written into metrics inside `_run_single_scenario`, which requires mocking subprocess calls, a running server, CSV parsing, and a MetricsProcessor. Other parameters that follow the exact same pattern (`architecture`, `repository`, `io_threads`, `benchmark_threads`) also don't have runner-level tests for the same reason. What the current design already tested is the postgres layer. It verifies the schema detection gives it VARCHAR(50), the value appears correctly in converted rows, and it truncates properly if over max length.

Carol-ggg:
Add cluster_mode column and related tests in 7c069f42088168a9175321c78a7c787eb657dfaf. Add skip_config support in 3f2a3a4319b4e46ac2b33020ba448c2faa445d28

Carol-ggg:
Update the grafana format to match the last grafana json commit 1e27ce0d4874fe651caa938497cf7ce72ed9a187

Carol-ggg:
Added function to retrospectively mark subset and relevant tests. changes in c3a1600747125a123335967a3dbedd89d4f96119

utils/module_postgres_track_commits.py
VoletiRam: Discussed offline that it is good to raise the error since we don't expect to run two workflows at the same time with same config pushing to same infrastructure.

utils/module_postgres_track_commits.py
```diff
@@ -0,0 +1,1129 @@
...
+        module_name: Module name (determines table)
+        config_name: Config file name to match
+        config_sets: List of module runtime configs
+        architecture: Architecture to match
+        cluster_mode: List of cluster mode booleans to match
+        max_pairs: Maximum number of pairs to fetch
+
+    Returns:
+        Tuple of (pairs, max_timestamps) where:
+            pairs: List of 'core_sha:module_sha' strings marked as in_progress
```
VoletiRam, line 583: Do we need to update this comment as it is returning list?

utils/module_postgres_track_commits.py
```diff
@@ -0,0 +1,1129 @@
...
+                     max_commit_timestamp DESC, min_commit_timestamp DESC
+            LIMIT %s
+        """,
+            (config_name, config_sets_json, architecture, cluster_mode_json, max_pairs),
+        )
+        rows = cur.fetchall()
+
+        if not rows:
+            print("No pending pairs to fetch", file=sys.stderr)
+            return [], []
```
VoletiRam, line 607: Here we are returning tuple. Can we fix the return type annotations properly?

utils/module_postgres_track_commits.py
```diff
@@ -0,0 +1,1129 @@
...
+                    UPDATE {table}
+                    SET status = 'completed_as_subset', priority = 99, updated_at = NOW()
+                    WHERE id = ANY(%s)
+                """,
+                    (ids_to_mark,),
+                )
+                total_marked += cur.rowcount
+
+    if total_marked > 0:
+        conn.commit()
```
VoletiRam: We are committing inside here but the caller doesn't know. Also, mark_module_commits already commits before calling this method.

The function does a commit inside the loop per-pair. If the connection drops between the first commit (marking done) and the second commit (retroactive subset), you get an inconsistent state. Both should be in one transaction.

__Fix:__ Remove `conn.commit()` from `_retroactively_mark_subsets`. Let the caller manage the transaction boundary, or combine both operations under one commit.

roshkhatri (commented):
We need to decide if we need the 2D tracking array, I think it is an overkill for the changes that we are doing, and then while visualizing it would be difficult as well, can you also post some screenshots in the top comment of the dashboards?

I think it will be better if we can just track the module commits and also enter the core commits in the table. If and when we see unreasonable results, we can check the core dashboard and compare

On another node, can you add the benchmark workflow here in the PR so, all the benchmarking and dashboarding code stays here. We can house all the perf benchmarking code in this repository

utils/module_postgres_track_commits.py
Carol-ggg: The current design already satisfied that. 
If there is a conflict, PostgreSQL raises a psycopg2.IntegrityError, the whole batch insert is rollbacked and then raised a value-error. 
But I have updated the error message to be more descriptive.

utils/module_postgres_track_commits.py
Carol-ggg: It is actually returning a tuple for both pairs and max_timestamp. I have updated the return type annotation to avoid any confusion

utils/module_postgres_track_commits.py
Carol-ggg: This is the same function as the previous comment. The return type annotation is updated.

utils/module_postgres_track_commits.py
Carol-ggg: This is removed. All the changes are in f0a1038dac54640dc47197d0e8288432375f352e

PR #59, opened 2026-06-17
Add Grafana dashboard GitOps sync via gcx + enable kubernetesDashboards

roshkhatri:
- Add .github/workflows/sync-dashboards.yml: syncs changed dashboards/grafana/*.json to Grafana on push to main using the Grafana CLI (gcx). Converts bare dashboard JSON to gcx resource manifests; validates and dry-runs before pushing.
- Guard step: workflow only runs when GRAFANA_API_TOKEN secret is set (no-op on forks / before setup).
- Snapshot live dashboards into the repo: valkey-performance-per-command, valkey-throughput-comparison, valkey-memory-overhead (replaces stale valkey-dashboard-public.json).
- grafana-values.yaml: root_url -> perf-dashboard.valkey.io, datasource connMaxLifetime=300 (avoid 350s NAT/ALB idle drops), move database into jsonData (Grafana v12+ requirement).


dashboards/grafana/grafana-values.yaml
```diff
@@ -68,7 +72,7 @@ grafana.ini:
     ssl_mode: require
     max_idle_conn: 2
     max_open_conn: 10
-    conn_max_lifetime: 14400
+    conn_max_lifetime: 0
```
rainsupreme: My pal Claude says this means "never expire connections". Should this be 300 or non-zero to match connMaxLifetime in the datasource below?

.github/workflows/sync-dashboards.yml
```diff
@@ -0,0 +1,120 @@
...
+      - name: Checkout repository
+        if: steps.guard.outputs.run == 'true'
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 2  # need the previous commit to diff against
+
+      - name: Install gcx
+        if: steps.guard.outputs.run == 'true'
+        run: |
+          curl -fsSL https://raw.githubusercontent.com/grafana/gcx/main/scripts/install.sh \
```
rainsupreme: nit: pin this to a specific release or commit sha? vulnerable to upstream breaking change or supply chain attack

dashboards/grafana/valkey-memory-overhead.json
```diff
@@ -0,0 +1,174 @@
...
+  "editable": true,
+  "fiscalYearStartMonth": 0,
+  "graphTooltip": 0,
+  "id": 5,
+  "links": [],
+  "panels": [
+    {
+      "datasource": {
+        "type": "grafana-postgresql-datasource",
+        "uid": "PDA4518EDFA6642B8"
```
rainsupreme: nit: this is defined in 3 different places and looks generated. If grafana rebuilt or re-provisioned this would change and break the dashboard. Could use a stable option

rainsupreme (approved):
Only real concern is conn_max_lifetime set to 0 - please double check that. Otherwise just a few nits.

I'm going on Madelyn's opinion that this isn't production code and has a lower bar, thus we can lean towards higher velocity, faster iteration, and be willing to potentially break things and quickly revert/fix. The perf dashboard has a lot of potential for improvement! 😁

PR #60, opened 2026-07-06
Fix benchmark comparison metric handling and significance logic

michellee-10:
Remove redundant latency_ prefix mapping that made the comparison pipeline fragile. All metrics now use canonical field names (avg_latency_ms, p50_latency_ms, etc.) consistently without fallbacks.

Fix _get_significance_indicator to correctly distinguish latency (lower-is-better) from throughput (higher-is-better) when determining if a change is an improvement or regression.

Add end-to-end pipeline tests covering both changes.

### DIFF
--- a/utils/compare_benchmark_results.py
+++ b/utils/compare_benchmark_results.py
@@ -924,7 +924,7 @@ def _generate_summary(
                 row.get("new_ci_lower", 0.0),
                 row.get("new_ci_upper", 0.0),
                 row["change"],
-                row.get("metric", ""),
+                row["metric"],
             )
 
             test_label = f"{row['command']} {row['metric']} pipe={row['pipeline']} threads={row['io_threads']}"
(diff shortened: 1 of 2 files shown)


utils/compare_benchmark_results.py
```diff
@@ -1218,13 +1184,16 @@ def _get_significance_indicator(
     if baseline_run_count <= 1 or new_run_count <= 1:
         return "❔"
 
-    # Check CI overlap
-    # No overlap if new's lower bound > baseline's upper bound (improvement)
-    # or new's upper bound < baseline's lower bound (regression)
+    # For latency metrics, lower values are better (less delay)
+    # For throughput metrics (rps), higher values are better
+    lower_is_better = "latency" in metric
```
roshkhatri, line 1189: Yeah, I was going to bring this up, thanks for addressing that.

utils/compare_benchmark_results.py
```diff
@@ -1204,6 +1169,7 @@ def _get_significance_indicator(
     new_ci_lower: float,
     new_ci_upper: float,
     change_percent: float,
+    metric: str = "",
```
roshkhatri: Can we make the `metric` field mandatory? If a future caller omits it, it will hit the bug again?

roshkhatri (commented):
The Pr looks good, just one minor change requested

roshkhatri (approved):
LGTM, Thanks @michellee-10 for resolving this

PR #61, opened 2026-07-07
Add mixed workload scenario and fuzzy dataset transform

VoletiRam:
Add two capabilities for scenario-based (test_groups) configs:

* type: mixed - spawn concurrent write and read sub-scenarios in the same test window. Each sub-scenario runs as its own valkey-benchmark process on its own CPU range and produces its own metric entry. options on a mixed scenario apply to each read sub-scenario.

* fuzzy transform + fuzzy/tag_only query types - deterministic base words per term_id with target_distance Levenshtein-edit variants. Query generator uses the same seed so every query has known matches in the dataset. Fixes a latent 'import random' shadowing bug in the existing numeric_range / tag_list branches.

Also adds README docs and focused unit tests for both features.

### DIFF
--- a/scripts/setup_datasets.py
+++ b/scripts/setup_datasets.py
@@ -3,6 +3,7 @@
 
 import argparse
 import csv
+import hashlib
 import json
 import logging
 import random
@@ -11,7 +12,21 @@ import sys
 import urllib.request
 import xml.etree.ElementTree as ET
 from pathlib import Path
+
 import numpy as np
+
+
+def _make_rng(filename: str) -> "np.random.Generator":
+    """Return a numpy Generator seeded deterministically from `filename`.
+
+    Two invocations against the same filename produce byte-identical output,
+    so vector datasets and query vectors are reproducible across machines.
+    Different filenames get different content (avoids accidentally emitting
+    the same random values for e.g. hybrid data and queries).
+    """
+    seed = int(hashlib.sha256(filename.encode("utf-8")).hexdigest()[:8], 16)
+    return np.random.default_rng(seed)
+
 
 # Alphabet used for deterministic random-word generation (fuzzy datasets)
 ALPHABET = "abcdefghijklmnopqrstuvwxyz"
@@ -226,12 +241,16 @@ def apply_transforms(
 
         elif ttype == "fuzzy":
             # Generate fuzzy-match dataset: multiple misspelling variants per base term.
-            # Structure mirrors "expansion": variant_count × docs_per_variant × term_count.
-            # Variant 0 is the correctly-spelled base word; other variants apply
+            # Structure: variant_count × docs_per_variant. Variant 0 is the
+            # correctly-spelled base word; other variants apply
             # ``target_distance`` Levenshtein edits (insert/delete/substitute).
+            #
+            # ``min_word_length`` / ``max_word_length`` / ``target_distance``
+            # MUST match the corresponding fuzzy query-generator config so that
+            # each query resolves to the same base word its dataset variant 0
+            # holds. See TestGenerateFuzzyQueries.
             variant_count = t.get("variant_count", 5)
             docs_per_variant = t.get("docs_per_variant", 20)
-            term_count = t.get("term_count", 100)
             min_word_length = t.get("min_word_length", 5)
             max_word_length = t.get("max_word_length", 6)
             target_distance = t.get("target_distance", 1)
@@ -252,11 +271,9 @@ def apply_transforms(
                 # First variant is the correctly-spelled base word (matches queries)
                 variant = base_word
             else:
-                # Deterministic per-(term_id, variant_id) seed. We combine the two
-                # into a single integer (safe for variant_count up to 1_000_000)
-                # so the RNG is stable across Python versions - hash() of tuples
-                # is deterministic within one process but its algorithm is an
-                # implementation detail.
+                # Deterministic seed: term_id * 1_000_000 + variant_id
+                # (assumes variant_id < 1_000_000, which is trivially true for
+                # realistic ``variant_count`` values).
                 variant_seed = term_id * 1_000_000 + variant_id
                 variant_rng = random.Random(variant_seed)
                 variant = base_word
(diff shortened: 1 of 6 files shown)


roshkhatri (commented):
We have merged features for the mixed workload scenario (valkey_benchmark.py), the fuzzy transform + query generators, and the vector NPY generation (setup_datasets.py) all in one PR, I think splitting them into smaller PRs would be better to track. WDYT?

scripts/setup_datasets.py
```diff
@@ -210,25 +293,144 @@ def apply_transforms(
...
+    # Fill each field. All inner loops are numpy-vectorized so throughput scales
+    # linearly with doc_count instead of degrading at 10^5+ rows.
+    for field in field_configs:
+        field_name = field["name"]
+        for t in field.get("transforms", []):
+            ttype = t.get("type")
+
+            if ttype == "vector":
+                dims = t.get("dimensions", 256)
+                vectors = np.random.randn(doc_count, dims).astype(np.float32)
```
roshkhatri: This uses the global numpy RNG with no seed (also lines 418, 426, and 725 in the query generator), so vector datasets and query vectors are not reproducible. The `output.exists()` cache makes it stable on one host, but two machines generating from the same config get different workloads, which breaks cross-machine comparisons. This also contradicts the fuzzy transform in this same PR, where determinism is carefully engineered and tested. Please use `np.random.default_rng(seed)` with a fixed or config-supplied seed.

scripts/setup_datasets.py
```diff
@@ -457,6 +669,69 @@ def generate_queries(output_dir: Path, config: dict, filename: str) -> Path:
             for term_id in range(1, num_queries + 1):
                 writer.writerow([f"term{term_id:03d}"])
 
+        elif query_type == "fuzzy":
+            # Generate the correctly-spelled base words for fuzzy datasets.
+            # Uses the same term_id → base_word mapping as the fuzzy transform,
+            # so each query is guaranteed to have matching variants in the dataset.
+            min_length = config.get("min_word_length", 5)
```
roshkhatri, line 711: The query side derives base words from `min_word_length`/`max_word_length` in the query config, while the dataset side derives them from the transform config, both via `random.Random(term_id)`. If the two ever diverge, every query matches zero rows and the benchmark quietly reports meaningless numbers. Can we validate at generation time that these match the fuzzy transform's params, or derive both from a shared block? At minimum the README should say they MUST match.

roshkhatri (commented):
Thanks for the PR! The mixed workload plumbing is clean and the determinism contracts pinned down in the fuzzy tests are exactly right. Main concerns are the unseeded vector generation and the fuzzy query/dataset config coupling; the rest are minor. 

*Reviewed with AI assistance.*

scripts/setup_datasets.py
```diff
@@ -210,25 +293,144 @@ def apply_transforms(
...
+        field_name = field["name"]
+        for t in field.get("transforms", []):
+            ttype = t.get("type")
+
+            if ttype == "vector":
+                dims = t.get("dimensions", 256)
+                vectors = np.random.randn(doc_count, dims).astype(np.float32)
+                data[field_name] = _l2_normalize(vectors)
+
+            elif ttype == "proximity_phrase":
```
roshkhatri: This reuses the `proximity_phrase` name but ignores `term_count`/`combinations` and emits a single token `phrase{qid}`, so it isn't a proximity phrase. The alignment test even passes those parameters and they do nothing. Either honor the parameters or give this its own name (`phrase_label`?) so the next config author isn't misled.

scripts/setup_datasets.py
```diff
@@ -183,6 +224,48 @@ def apply_transforms(
...
                 content = " ".join(parts)
 
+        elif ttype == "fuzzy":
+            # Generate fuzzy-match dataset: multiple misspelling variants per base term.
+            # Structure mirrors "expansion": variant_count × docs_per_variant × term_count.
+            # Variant 0 is the correctly-spelled base word; other variants apply
+            # ``target_distance`` Levenshtein edits (insert/delete/substitute).
+            variant_count = t.get("variant_count", 5)
+            docs_per_variant = t.get("docs_per_variant", 20)
+            term_count = t.get("term_count", 100)
```
roshkhatri: `term_count` is read here but never used; `term_id` is derived purely from `doc_num // docs_per_term`. Either use it to cap/wrap `term_id` or drop it, so the config surface doesn't imply a knob that does nothing.

scripts/setup_datasets.py
```diff
@@ -183,6 +224,48 @@ def apply_transforms(
...
+            if variant_id == 0:
+                # First variant is the correctly-spelled base word (matches queries)
+                variant = base_word
+            else:
+                # Deterministic per-(term_id, variant_id) seed. We combine the two
+                # into a single integer (safe for variant_count up to 1_000_000)
+                # so the RNG is stable across Python versions - hash() of tuples
+                # is deterministic within one process but its algorithm is an
+                # implementation detail.
+                variant_seed = term_id * 1_000_000 + variant_id
```
roshkhatri, line 277: nit: the comment describes a `hash()`-of-tuples approach the code doesn't take, and the "safe for variant_count up to 1_000_000" bound is actually on `variant_id`. Trimming it to "deterministic seed = term_id * 1_000_000 + variant_id" would be clearer.

tests/test_setup_datasets.py
```diff
@@ -0,0 +1,258 @@
...
+    side changes format (say, ``phrase_{i}`` or ``phrase-{i}``), Group 15 will
+    silently return zero results and the whole KNN + text-prefilter benchmark
+    becomes meaningless. This is the same class of fragile contract that
+    ``TestGenerateFuzzyQueries`` guards for fuzzy.
+    """
+
+    def test_every_query_term_appears_in_the_hybrid_dataset(self, tmp_path: Path):
+        try:
+            import numpy as np
+        except ImportError:  # pragma: no cover - numpy is a hard dep
```
roshkhatri: This skip guard is dead code: `import setup_datasets` at module scope already imports numpy, so a missing numpy fails the file at collection before the skip can run. numpy is pinned in requirements.txt, so this can just be dropped.

valkey_benchmark.py
```diff
@@ -987,6 +1033,270 @@ def _execute_setup_command(self, cmd_str: str) -> None:
...
+        # Determine target ports based on cluster execution mode
+        if scenario.get("cluster_execution") == "single" or not self._is_cme():
+            ports = [self._get_active_ports()[0]]
+        else:
+            ports = self._get_active_ports()
+
+        total_writes = sum(w.get("clients", 1) for w in write_scenarios)
+        total_reads = sum(r.get("clients", 1) for r in read_scenarios)
+        logging.info(
+            f"Mixed workload: {total_writes} write clients + {total_reads} read clients "
```
roshkhatri, line 1234: nit: in CME parallel mode each sub-scenario launches one process per node, so the real client total is `sum(clients) * len(ports)`, not `sum(clients)`. The log undercounts.

tests/test_client_runner_logic.py
```diff
@@ -518,3 +520,262 @@ def test_simple_format_has_combination_keys(self, minimal_client_runner):
...
+    def test_metrics_have_correct_test_id_and_phase_per_sub_scenario(
+        self, mixed_runner, mixed_scenario
+    ):
+        """Every produced metric must be attributable to its sub-scenario id."""
+        # Use side_effect (not return_value) so each call returns a *fresh*
+        # dict. _create_mixed_metric mutates the dict returned by
+        # create_metrics; a shared return_value would cause every metric to
+        # end up as the same dict with the last-written test_id.
+        metrics_processor = MagicMock()
+        metrics_processor.create_metrics.side_effect = lambda *a, **kw: {"rps": 1000.0}
```
roshkhatri, line 693: The lambda mock means the 8-positional-arg call in `_create_mixed_metric` is never checked against the real `MetricsProcessor.create_metrics` signature. One test with the real processor would catch a signature drift.

roshkhatri:
Summary of my review comments, to make prioritization easier:

| # | Comment | File | Priority | Confidence |
|---|---------|------|----------|------------|
| 1 | Unseeded vector generation (reproducibility) | `scripts/setup_datasets.py` | Blocking | High |
| 2 | Fuzzy query/dataset config coupling | `scripts/setup_datasets.py` | Blocking | High |
| 3 | `proximity_phrase` NPY path ignores its params | `scripts/setup_datasets.py` | Should fix | High |
| 4 | Dead `term_count` in fuzzy transform | `scripts/setup_datasets.py` | Should fix | High |
| 5 | Misleading `variant_seed` comment | `scripts/setup_datasets.py` | Nit | High |
| 6 | Dead numpy skip guard in test | `tests/test_setup_datasets.py` | Nit | Medium |
| 7 | Log undercounts clients in CME mode | `valkey_benchmark.py` | Nit | Medium |
| 8 | Lambda mock skips real `create_metrics` signature check | `tests/test_client_runner_logic.py` | Optional | Medium |

roshkhatri (approved):
Looked into the comments resolution and the code, I think This looks good

PR #62, opened 2026-07-08
Adds module commit tracking support for valkey modules

Carol-ggg:
## Summary

Adds module-level commit tracking for valkey module benchmarks by extending the existing tracking scripts.

**`postgres_track_commits.py`**
- New `--module-name` flag routes tracking to a per-module table (`benchmark_module_commits_<name>`), separate from core's `benchmark_commits`.
- Config file is loaded with runtime overrides (`--cluster-mode-filter`, `--skip-config-set`, `--skip-profiling`) so tracking matches what's actually benchmarked.
- For module tracking, large config keys (`test_groups`, `dataset_generation`, `query_generation`) are stripped and a `config_name` is derived from the config filename and stored.
- Cleanup can now scope to a specific config+architecture instead of always wiping all in-progress rows (needed for concurrent module runs).

**`push_to_postgres.py`**
- New `--module-commit-timestamp` stores the module commit's time in its own column, keeping the core commit `timestamp` intact.
- Table name auto-generates from `--module` when `--table-name` isn't given.
- Schema handling added for `config_name`, `config_set`, `module_commit` fields.

**`benchmark.py` / `valkey_benchmark.py`**
- `config_name` (from the config filename) is threaded through `benchmark.py` into `ClientRunner`, which writes it into each `metrics.json` result for module runs.

**`.github/workflows/search_benchmark.yml`**
- Switched from tracking core+module commit pairs to tracking module commits only; core is always pinned to the latest `unstable` HEAD (or a manual override).
- Added manual override inputs (`core_commit`, `module_commits`) for ad-hoc runs.
- Calls the shared `postgres_track_commits.py` script with `--module-name` instead of a separate module-specific script.

**New config/dashboard files**
- `configs/fts-benchmarks-arm.json` and `dashboards/grafana/valkey-module-perf.json` added and updated to fit the current design.

**Tests**
- Updated tests to cover all changes.

Screenshot 2026-07-08 at 3 14 09 PM
Screenshot 2026-07-08 at 3 15 04 PM


configs/fts-benchmarks-arm.json
VoletiRam, line 1: why do we move this json file here?

utils/postgres_track_commits.py
```diff
@@ -13,12 +13,156 @@
...
+
+    Raises:
+        ValueError: If module_name is empty string (likely accidental omission).
+    """
+    if module_name is None:
+        return CORE_TABLE_NAME
+    if not module_name.strip():
+        raise ValueError("Module name cannot be empty.")
+    return f"benchmark_module_commits_{module_name}"
+
```
VoletiRam: Did we revert the regex validation to prevent sql injection?

Can we restore the regex from PR #58:

```python
if not re.match(r"^[a-z][a-z0-9_]{0,30}$", module_name):
    raise ValueError(f"Invalid module_name: {module_name}")
```

benchmark.py
```diff
@@ -815,6 +840,9 @@ def main() -> None:
                 args=args,
                 module_path=module_path,
                 uses_test_groups=uses_test_groups,
+                config_name=Path(args.config).name if args.module else None,
```
VoletiRam, line 852: Path(args.config).name` crashes when `--module` set without `--config`

```python
# benchmark.py line 843
config_name=Path(args.config).name if args.module else None,
```

If someone runs with `--module search` but no `--config`, `args.config` is `None`, and `Path(None)` raises `TypeError`. Should be:

```python
config_name=Path(args.config).name if args.module and args.config else None,
```

.github/workflows/search_benchmark.yml
```diff
@@ -0,0 +1,395 @@
...
+      CORE_DIR: "valkey"
+      CORE_PATH: "${{ github.workspace }}/valkey"
+      # Benchmark configuration
+      PERF_BENCHMARK_REPO: "valkey-io/valkey-perf-benchmark"
+      PERF_BENCHMARK_BRANCH: "main"
+      PERF_BENCHMARK_DIR: "valkey-perf-benchmark"
+      PERF_BENCHMARK_PATH: "${{ github.workspace }}/valkey-perf-benchmark"
+      CONFIG_FILE: "${{ github.workspace }}/valkey-perf-benchmark/configs/fts-benchmarks-arm.json"
+      # Run parameters
+      NUM_RUN: ${{ inputs.num_runs || '5' }}
```
VoletiRam: We give input default 1 above(default: "1" ). Can we make it consistent? Does num_runs have any effect on module tests?

utils/postgres_track_commits.py
```diff
@@ -13,12 +13,156 @@
...
+        skip_profiling: If True, sets 'profiling_sets' to default [{"enabled": False}].
+
+    Returns:
+        Modified config dict.
+    """
+    if cluster_mode is not None:
+        if cluster_mode.lower() == "true":
+            cfg["cluster_mode"] = True
+        elif cluster_mode.lower() == "false":
+            cfg["cluster_mode"] = False
```
VoletiRam, line 76: We are mutates AND returns in this `_apply_config_overrides`. Pick one. Either:

- Mutate in-place → return None
- Return a copy → don't mutate

The list-comprehension usage `[_apply_config_overrides(cfg, ...) for cfg in config]` reads as "transform" but actually mutates every element of the caller's list.

utils/postgres_track_commits.py
```diff
@@ -13,12 +13,156 @@
...
+        if isinstance(config, dict):
+            config = {k: v for k, v in config.items() if k not in skip_keys}
+            config["config_name"] = config_name
+        elif isinstance(config, list):
+            config = [
+                {k: v for k, v in cfg.items() if k not in skip_keys}
+                for cfg in config
+                if isinstance(cfg, dict)
+            ]
+            config.insert(0, {"config_name": config_name})
```
VoletiRam, line 132: hmm. The list of "config sets" now has a header dict `{"config_name": "..."}` that isn't a config set. Downstream consumers (`_is_config_array_subset`) will try to match this against real configs during subset detection.


Better add `config_name` as a scalar column in the table, or add it once to each dict:

```python
config = [{**cfg_stripped, "config_name": config_name} for cfg in config]
```

utils/postgres_track_commits.py
```diff
@@ -13,12 +13,156 @@
...
+        if cluster_mode.lower() == "true":
+            cfg["cluster_mode"] = True
+        elif cluster_mode.lower() == "false":
+            cfg["cluster_mode"] = False
+    if skip_config_set:
+        cfg["config_sets"] = DEFAULT_CONFIG_SETS
+    if skip_profiling:
+        if "profiling_sets" not in cfg:
+            print(
+                "Warning: --skip-profiling passed but no 'profiling_sets' in config",
```
VoletiRam: This warning is not necessary. If they opted to skip profiling, we shouldn't worry about presence of perf related configs.

utils/postgres_track_commits.py
```diff
@@ -423,23 +634,24 @@ def get_commits_by_config(
...
-    create_tables(conn)
+    create_tables(conn, table_name)
 
     with conn.cursor() as cur:
         cur.execute(
-            """
+            f"""
             SELECT DISTINCT config
-            FROM benchmark_commits
+            FROM {table_name}
```
VoletiRam, line 637: Noticing it now. Do we need to filter this by architecture?

Cross-architecture data mixed together in the output. Since arch is a legitimate filter dimension, this should accept an optional arch filter i feel.

utils/postgres_track_commits.py
VoletiRam, line 1: Create_tables()` called from EVERY entry-point function (5x)

```python
def cleanup_incomplete_commits(...):
    create_tables(conn, table_name)  # <-- here
def determine_commits_to_benchmark(...):
    create_tables(conn, table_name)  # <-- here
def get_commits_by_config(...):
    create_tables(conn, table_name)  # <-- here
```

Not a bug but `CREATE TABLE IF NOT EXISTS` is idempotent. 5 CREATE + 5×5 CREATE INDEX = 30 no-op queries per invocation. Move to `main()` after DB connection. PR #58 did this correctly;

tests/test_postgres_utils.py
VoletiRam, line 1: `TestBuildCleanupQuery.test_no_filters` and `test_architecture_without_config_is_ignored` produce the same output. Can we dedup them.

VoletiRam, line 1: Can we add a test case for sql injection to test `_resolve_module_table_name`?

VoletiRam, line 1: Missing test coverage for _apply_config_overrides, _find_superset_configs, workflow default vs input default mismatch.


`test_module_commit_timestamp_parsed_to_datetime` is under `TestLoadConfig` but tests `convert_metrics_to_row.` Do we need to move it?

utils/postgres_track_commits.py
Carol-ggg: We can definitely moved it to `main()`, but this was the original design before my PR. Do you want to update this change and move the `create_tables` to `main()`?

configs/fts-benchmarks-arm.json
Carol-ggg: As we moved the workflow to this repo, I also updated the test.json file to be here also. Would you prefer to keep the the test json in the search repo?

benchmark.py
Carol-ggg: In the original design config is default to `default="./configs/benchmark-configs.json` if not provided. `main() `also load_config directly (so it will break before it reaches this line if config is not provided)
But we could add a warning if config is not provided and fall back to default for logging purposes. Would you want me to add that?

utils/postgres_track_commits.py
Carol-ggg: If all module testing follow this pattern, then the subset would correctly detect two config json with the same `config_name` insert, so this wouldn't actually affect the `subset_detection` result. 

1. for adding as a scalar column, this would lead to a lot of changes in the downstream sql queries and I think it might not be the most elegant change. 
2. If a config_file is a list of multiple dict, it will still be stored on one row. I think it would be unnecessary repeat to insert the name into each dict.

utils/postgres_track_commits.py
Carol-ggg: This is a pre-existing design, currently it is not directly used in the workflow and only exist as a debugging tool. But I can definitely add an optional architecture filter to this. Would you want to implement this change here?

.github/workflows/search_benchmark.yml
Carol-ggg: Currently the num_runs is not applicable for module test. I have removed it for now and we can add it back if we later can support multiple runs. changes in b00e09bfe6d5cf2fe867279c31b5eaf1546cd8c5

utils/postgres_track_commits.py
Carol-ggg: Sorry for missing that. I have updated the logic for both `postgres_track_commits` and the one is `push_to_postgres`. changes in b00e09bfe6d5cf2fe867279c31b5eaf1546cd8c5

utils/postgres_track_commits.py
Carol-ggg: That's a good catch. Thanks so much. I changed the return to None and indicates that it replace in place. I have also changed the list comprehension to a for loop so it would be more readable. changes in b00e09bfe6d5cf2fe867279c31b5eaf1546cd8c5

utils/postgres_track_commits.py
Carol-ggg: Make sense, I will remove that warning. changes in b00e09bfe6d5cf2fe867279c31b5eaf1546cd8c5

tests/test_postgres_utils.py
Carol-ggg: They tested different invariant, but you are right that since architecture is always auto-detected if not provided, I deleted the `test_no_filters` as this might not happen in real testing. changes in b00e09bfe6d5cf2fe867279c31b5eaf1546cd8c5

Carol-ggg: I added the sql injection test for both tables creation functions. changes in b00e09bfe6d5cf2fe867279c31b5eaf1546cd8c5

Carol-ggg: 1. `_apply_config_overrides` is not explicitly tested but is tested indirectly in `test_cluster_mode_override_true` and other related tesets under `TestLoadConfig`. Would you prefer me to add seperate unit tests? 
2. `_find_superset_configs` would need a SQL integration tests. 
3. I have moved both timestamp testing to more applicable places.

changes in b00e09bfe6d5cf2fe867279c31b5eaf1546cd8c5

utils/postgres_track_commits.py
roshkhatri: Yes you can move it

configs/fts-benchmarks-arm.json
VoletiRam: We previously moved the test json in respective repo as per @roshkhatri suggestion. I am okay either way. Will leave the decision to Roshan.

benchmark.py
VoletiRam: sure, but if it is expected to be handled in main(), we shouldn't have this check elsewhere, we just need to pass it down. Are we erroring in main if config is not provided?

utils/postgres_track_commits.py
VoletiRam: Depends on the usage while debugging. Do you see value adding it?

tests/test_postgres_utils.py
VoletiRam: > apply_config_overrides is not explicitly tested but is tested indirectly in test_cluster_mode_override_true and other related tesets under TestLoadConfig. Would you prefer me to add seperate unit tests?

if it is already covered by higher level, it is fine. If we are missing any edge cases, lets add the coverage.

> _find_superset_configs would need a SQL integration tests.

Okay, fine.

utils/postgres_track_commits.py
VoletiRam: i agree.

VoletiRam (approved):

benchmark.py
Carol-ggg: we are not erroring in `main()` now. it will just be the default `default="./configs/benchmark-configs.json`
"But we could add a warning if config is not provided and fall back to default for logging purposes. Would you want me to add that?" For this, I am referring to adding an warning in `main()`

utils/postgres_track_commits.py
Carol-ggg: Sound good. I will keep the current design then.

utils/postgres_track_commits.py
Carol-ggg: Depends on if we are testing different config on different architecture. 
But either way, the detailed information should all be available in the dropdown part of the grafana, which I think might be easier to check?

configs/fts-benchmarks-arm.json
roshkhatri: Yes, We will run manage all workflows from this repository, so the management of runners and the configs and action identities are contained

benchmark.py
roshkhatri: I think we should add a warning saying that we are choosing the above file as default

utils/postgres_track_commits.py
```diff
@@ -13,12 +14,147 @@
 from psycopg2.extras import Json
 
 
-def create_tables(conn):
-    """Create benchmark tracking tables if they don't exist."""
+CORE_TABLE_NAME = "benchmark_commits"
```
roshkhatri, line 17: Can we make core as the module names "core"? so we wouldnot need this, and we would pass core, or default the --module to core if --module is not passes. and then it can be dynamic

.github/workflows/search_benchmark.yml
```diff
@@ -0,0 +1,388 @@
...
+      - name: Set up Python
+        uses: kishaningithub/setup-python-amazon-linux@63dc7bd642c6a564d0a93dd4b1cbe3e448ce3621
+        with:
+          python-version: "3.10"
+          cache: "pip"
+
+      - name: Checkout valkey_ram (for dataset-enabled valkey-benchmark)
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        with:
+          repository: VoletiRam/valkey_ram
```
roshkhatri, line 78: Are we planning on adding this to the core? and using core for benchmark?

.github/workflows/search_benchmark.yml
```diff
@@ -0,0 +1,388 @@
...
+      # =========================================================================
+      # Commit tracking: determine module commits to benchmark
+      # =========================================================================
+
+      - name: Resolve latest core commit
+        id: core_commit
+        working-directory: ${{ env.CORE_DIR }}
+        run: |
+          if [[ -n "${{ inputs.core_commit }}" ]]; then
+            CORE_SHA="${{ inputs.core_commit }}"
```
roshkhatri: Can you add these as env. variables here? like `CORE_COMMIT_INPUT`

.github/workflows/search_benchmark.yml
```diff
@@ -0,0 +1,388 @@
...
+      PERF_BENCHMARK_DIR: "valkey-perf-benchmark"
+      PERF_BENCHMARK_PATH: "${{ github.workspace }}/valkey-perf-benchmark"
+      CONFIG_FILE: "${{ github.workspace }}/valkey-perf-benchmark/configs/fts-benchmarks-arm.json"
+      # Run parameters
+      MAX_MODULE_COMMITS: ${{ inputs.max_module_commits || '3' }}
+      CLUSTER_MODE: ""
+      SKIP_CONFIG_SET: ""
+      SKIP_PROFILING: ""
+    runs-on: [self-hosted, valkey-search-arm]
+    timeout-minutes: 7200
```
roshkhatri: Lets keep this lower closer to that actual time taken

.github/workflows/search_benchmark.yml
```diff
@@ -0,0 +1,388 @@
...
+              openssl-devel \
+              systemd-devel \
+              bzip2-devel \
+              libffi-devel \
+              perf 
+          
+          gcc --version
+          g++ --version
+          
+          pip install -r requirements.txt
```
roshkhatri:
```suggestion
          pip install --require-hashes -r requirements.txt
```

utils/postgres_track_commits.py
```diff
@@ -27,21 +163,20 @@ def create_tables(conn):
                 architecture VARCHAR(50),
                 created_at TIMESTAMPTZ DEFAULT NOW(),
                 updated_at TIMESTAMPTZ DEFAULT NOW(),
-                
-                -- Unique constraint: same commit + config + architecture can only exist once
-                CONSTRAINT unique_sha_config_arch UNIQUE(sha, config, architecture)
+
+                -- Unique constraint: same commit + config + architecture can only exist once                
+                CONSTRAINT unique{prefix}sha_config_arch UNIQUE(sha, config, architecture)
             );
```
roshkhatri, line 170: added whitespaces, is it required?

roshkhatri (commented):
Some minor concerns, other tracking looks good, some comments needs to be resolved as @VoletiRam pointed out about `create_tables`

Also should we add integration tests for the Postgres DB? or should do in the future.

.github/workflows/search_benchmark.yml
VoletiRam: Yes but we need to switch existing tests to match with oss binary feature. The current limitation is OSS only support CSV file format. We have other formats support in my fork. I will create new PR updating it as follow up PR.

VoletiRam (approved):

benchmark.py
Carol-ggg: warning added in 744f5db0a44e97540f988bea595229c0bd0ab613

utils/postgres_track_commits.py
Carol-ggg: moved to main(). changes in 744f5db0a44e97540f988bea595229c0bd0ab613

.github/workflows/search_benchmark.yml
Carol-ggg: added as env. variable. changes in 744f5db0a44e97540f988bea595229c0bd0ab613

.github/workflows/search_benchmark.yml
Carol-ggg: changed to 1800 (30hr). Current workflow with profiling takes 20hr. changes in 744f5db0a44e97540f988bea595229c0bd0ab613

utils/postgres_track_commits.py
Carol-ggg: deleted the whitespace. changes in 744f5db0a44e97540f988bea595229c0bd0ab613

VoletiRam (approved):

roshkhatri (commented):
I want to request for a minor change.

This is the comparison of the core and the module,

I want to make it standard

Can you remove ` --table-name "benchmark_metrics" ` from  `push_to_postgres.py --table-name "benchmark_metrics" ` and default the module to core
It would bring standardization I believe? also why cant `benchmark_module_commits_search` be `benchmark_commits_search` so the prefixes can be same as core but you add suffix when `--module` is mentioned in the parameter

`--module` and `--module-name` inconsistency in both scripts`

Carol-ggg:
> Can you remove `--table-name "benchmark_metrics"` from  `push_to_postgres.py --table-name "benchmark_metrics" ` and default the module to core

1. change `benchmark_module_commits_search` to `benchmark_commits_search` for tracking table for module
2. use `--module` for both files instead of the inconsistency design
3. remove the `--table-name` CLI argument support in `push_to_postgres` and also remove this argument from the core workflow. (leave unchanged for `regression_detection`) 

changes in a221dc3925430e760605958a55cc5a7d01b104d5

dashboards/grafana/valkey-module-perf.json
```diff
@@ -0,0 +1,1066 @@
...
+        "sort": 0,
+        "type": "query"
+      },
+      {
+        "current": {},
+        "datasource": {
+          "type": "postgres",
+          "uid": "PDA4518EDFA6642B8"
+        },
+        "definition": "SELECT REPLACE('${table_name}', 'benchmark_metrics_', 'benchmark_module_commits_')",
```
roshkhatri: This needs to change I think

dashboards/grafana/valkey-module-perf.json
```diff
@@ -0,0 +1,1066 @@
...
+        "hide": 2,
+        "includeAll": false,
+        "label": "Tracking Table",
+        "multi": false,
+        "name": "tracking_table",
+        "options": [],
+        "query": "SELECT REPLACE('${table_name}', 'benchmark_metrics_', 'benchmark_module_commits_')",
```
roshkhatri: This needs to change I think

roshkhatri (commented):
We need to update this in `benchmark_release_tags.yml` too

dashboards/grafana/valkey-module-perf.json
Carol-ggg: updated.

dashboards/grafana/valkey-module-perf.json
Carol-ggg: updated

Carol-ggg:
> We need to update this in `benchmark_release_tags.yml` too

updated --module to --table, and hardcode table name for 'tag'

.github/workflows/search_benchmark.yml
```diff
@@ -0,0 +1,390 @@
...
+              exit 1
+            fi
+            echo "✓ valkey-search module build successful"
+
+            # --- Run benchmark ---
+            cd "$PERF_BENCHMARK_PATH"
+            RESULTS_DIR="results/${CORE_SHA}_${MODULE_SHA}"
+            echo "Running search benchmarks for core=$CORE_SHA module=$MODULE_SHA..."
+            python benchmark.py \
+              --table "$MODULE_NAME" \
```
roshkhatri: Why did we change this: ` --table "$MODULE_NAME" \`?
Should be module?

Carol-ggg: its updated. thanks for catching that

VoletiRam (approved):

roshkhatri (approved):
The only this remaining for followup is the using CPU allocation to 2 NUMA nodes for server and client. Can do as a followup

Everything else looks good to be merges

PR #63, opened 2026-07-15
Add section explaining how to interpret significance indicators

michellee-10:
Documents the meaning of % change with uncertainty ranges and the significance indicators (✅, ❌, ➖, ❔) used in benchmark comparison output.


roshkhatri (approved):
LGTM

PR #64, opened 2026-07-18
Add shortened test and update AWS authentication for module workflow

Carol-ggg:
### Workflow (`.github/workflows/search_benchmark.yml`)
- Switched Postgres auth from static `PG_PASSWORD` secret to RDS IAM auth
  tokens over OIDC. Secrets used: `AWS_REGION`, `AWS_ROLE_TO_ASSUME`,
  `POSTGRESQL_ENDPOINT`, `RDS_USERNAME`, `RDS_DATABASE` (same names already
  wired up for the core workflow).
- Added an in-step `reauth_aws` bash function so STS + RDS tokens can be
  refreshed between iterations of the benchmark loop (each iteration is
  typically longer than either TTL).
- Added a `skip_profiling` boolean workflow_dispatch input. Default is
  off (full profiling); scheduled runs always profile.
- Added `Ensure git is installed` guard to prevent `actions/checkout` from
  silently falling back to the tarball API on runners with a broken git
  install.
- Cleaned up `MODULE_BRANCH: ${{ inputs.branch || 'main' }}` → hardcoded
  to `"main"` since no `branch` input was ever declared.

### Config (`configs/fts-benchmarks-shortened-arm.json`)
- New shorter FTS scenario set (4 test groups, 15 scenarios) for faster
  benchmark iterations.
- Per-scenario profiling delays on the four write scenarios so profile
  captures land on steady-state ingestion rather than warm-up.


configs/fts-benchmarks-shortened-arm.json
```diff
@@ -0,0 +1,623 @@
...
+      "per_cpu_enabled": false
+    },
+    "profiling_sets": [
+      {
+        "enabled": false
+      },
+      {
+        "enabled": true,
+        "mode": "wall-time",
+        "sampling_freq": 999,
```
VoletiRam, line 587: Are we enabling profiling or disabling here? I see two configs here with same key(enabled) but different value.

configs/fts-benchmarks-shortened-arm.json
```diff
@@ -0,0 +1,623 @@
...
+          "--use-coordinator"
+        ]
+      }
+    ],
+    "config_sets": [
+      {
+        "search.reader-threads": 8,
+        "search.writer-threads": 8,
+        "search.async-fanout-threshold": 4
+      }
```
VoletiRam, line 574: Can we also add single threaded version too?

configs/fts-benchmarks-shortened-arm.json
```diff
@@ -0,0 +1,623 @@
...
+    ],
```
VoletiRam, line 423: Can we add single threaded too? Can this still finish within 4 hours?

.github/workflows/search_benchmark.yml
```diff
@@ -252,9 +275,47 @@ jobs:
 
       - name: Benchmark module commits
         if: steps.determine_commits.outputs.skip == 'false'
-        env:
-          PGPASSWORD: ${{ secrets.PG_PASSWORD }}
         run: |
+          # AWS STS session credentials expire after 1 hour by default, and RDS
+          # IAM auth tokens expire after 15 minutes. Since a single iteration
+          # (build + benchmark) can take much longer than either, we refresh
+          # both right before every DB call:
```
roshkhatri: We can shorten this a bit

.github/workflows/search_benchmark.yml
```diff
@@ -252,9 +275,47 @@ jobs:
...
+          }
+
+          gen_pgpassword() {
+            aws rds generate-db-auth-token \
+              --hostname ${{ secrets.POSTGRESQL_ENDPOINT }} \
+              --port 5432 \
+              --region ${{ secrets.AWS_REGION }} \
+              --username ${{ secrets.RDS_USERNAME }}
+          }
+
```
roshkhatri: Why are we not using the github action?

Also configure this when we are just using it. in the above code, we auth before building, which is not req, we should auth, just before we use it, like at line 153

This can be moved just before line 191 ` name: Cleanup stale rows from previous runs`
```
  - name: Configure AWS credentials
        uses: aws-actions/configure-aws-credentials@v4
        with:
          aws-region: ${{ secrets.AWS_REGION }}
          role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
```

.github/workflows/search_benchmark.yml
```diff
@@ -326,10 +391,10 @@ jobs:
...
               --status complete \
-              --host ${{ secrets.RDS_HOST }} \
+              --host ${{ secrets.POSTGRESQL_ENDPOINT }} \
               --port 5432 \
-              --database postgres \
-              --username ${{ secrets.PG_USERNAME }} \
+              --database ${{ secrets.RDS_DATABASE }} \
+              --username ${{ secrets.RDS_USERNAME }} \
               --password "$PGPASSWORD" \
               "$MODULE_SHA"
```
roshkhatri: Why are these not different steps like in the core benchmark wf?

roshkhatri (changes requested):

configs/fts-benchmarks-shortened-arm.json
Carol-ggg: this is same with the original test json design. with enabled:false and enabled:true. I will run the whole test once with profiling disabled and once with the profiling enabled.

configs/fts-benchmarks-shortened-arm.json
Carol-ggg: it will double the existing time. currently, the whole test takes 30min. If we add single threaded, it would be 1hr. With profiling, it will be 2hr. Then we can only do num_runs=2 if we want to finish around 4hr.

.github/workflows/search_benchmark.yml
Carol-ggg: Answer both this and the latter comment. 
Currently, we are looping over multiple module commits, so they are in one `run:` step. Because they are in one step, we cannot use the github action version for AWS credentials (we cannot use `uses:` in `run:` step). 
Code at 280 is just for shell function declaration, it is not called until it is actually used at line 370(for push to postgres) and line 380 (for mark complete).

configs/fts-benchmarks-shortened-arm.json
VoletiRam: Aah makes sense. I confused as its been a while. Thanks for clarifying.

.github/workflows/search_benchmark.yml
Carol-ggg: The one at 153 is moved to just before 193. changes in da7b3689b999181ec2da282ca767e57b7da7a8e5

.github/workflows/search_benchmark.yml
Carol-ggg: this is updated in da7b3689b999181ec2da282ca767e57b7da7a8e5

.github/workflows/search_benchmark.yml
roshkhatri: Yes, for multiple module commits, why do we need AWS creds to tun the benhcmark?
We run the benchmarks, write the results in json format, and then push all the metrics in one step

configs/fts-benchmarks-shortened-arm.json
VoletiRam: We don't want to include profiling for daily regression testing. We want profiling on demand based on PR. I am fine on-demand PR with profiling to take longer. I believe without profiling,  1 + 8 threads can still be done within 4 hours. Can you please run once to note the duration of overall with 3 reruns to confirm this?

.github/workflows/search_benchmark.yml
Carol-ggg: moved `push_to_postgres` and `mark complete` outside of the loop. use the action for AWS authentication. changes in a0ef019738ff29544dd28d0687f4579c87fd8c5f

configs/fts-benchmarks-shortened-arm.json
Carol-ggg: single threaded config set added. profiling default to skip. also added one yml file for full test json (no cron on demand only). changed in ea3ff571cfe07adb6203b24d68baa82080dbf68a

VoletiRam (approved):

.github/workflows/search_benchmark.yml
```diff
@@ -55,17 +60,20 @@ jobs:
...
-      CONFIG_FILE: "${{ github.workspace }}/valkey-perf-benchmark/configs/fts-benchmarks-arm.json"
+      CONFIG_FILE: "${{ github.workspace }}/valkey-perf-benchmark/configs/fts-benchmarks-shortened-arm.json"
       # Run parameters
       MAX_MODULE_COMMITS: ${{ inputs.max_module_commits || '3' }}
       CORE_COMMIT_INPUT: ${{ inputs.core_commit || '' }}
       MODULE_COMMITS_INPUT: ${{ inputs.module_commits || '' }}
       CLUSTER_MODE: ""
       SKIP_CONFIG_SET: ""
-      SKIP_PROFILING: ""
+      SKIP_PROFILING: ${{ (inputs.skip_profiling == false) && '' || '1' }}
```
roshkhatri: why not just keep true and false?

Carol-ggg: `${SKIP_PROFILING:+--skip-profiling` only checks if it s an empty string or not, it does not evaluate boolean

roshkhatri (commented):
Lets merge the 2 workflows as they are the essentially the same  and decide the config file that we manually input, or for a cron we can use the one we want to run on a cron.
WDYT?

Carol-ggg:
> Lets merge the 2 workflows as they are the essentially the same and decide the config file that we manually input, or for a cron we can use the one we want to run on a cron. WDYT?

`search_benchmark_full.yml` removed. added a config-file choice for manual input. default/cron uses the `fts-benchmarks-shortened-arm.json`

.github/workflows/search_benchmark.yml
```diff
@@ -55,17 +79,20 @@ jobs:
...
-      MAX_MODULE_COMMITS: ${{ inputs.max_module_commits || '3' }}
+      MAX_MODULE_COMMITS: ${{ inputs.max_module_commits || '1' }}
       CORE_COMMIT_INPUT: ${{ inputs.core_commit || '' }}
       MODULE_COMMITS_INPUT: ${{ inputs.module_commits || '' }}
-      CLUSTER_MODE: ""
-      SKIP_CONFIG_SET: ""
-      SKIP_PROFILING: ""
+      CLUSTER_MODE: ${{ inputs.cluster_mode || '' }}
+      SKIP_CONFIG_SET: ${{ inputs.skip_config_set && '1' || '' }}
+      SKIP_PROFILING: ${{ (inputs.skip_profiling == false) && '' || '1' }}
```
roshkhatri:
```suggestion
      SKIP_PROFILING: ${{ (github.event_name == 'schedule' || inputs.skip_profiling) && '1' || '' }}
```
If we want to skip profiling on the scheduled runs this would be the correct condition right? 

I feel the condition rn might be incorrect

Carol-ggg: Thanks! What you suggested is better. I just double checked and realized that left side having '' would be seen as falsy and thus this would evaluate to right side regardlessly.

roshkhatri (approved):

PR #66, opened 2026-07-24
Add GCC installation in workflow for search module build 

Carol-ggg:
Adds an idempotent step to install GCC 14 from AL2023 repos if the runner doesn't already have GCC >= 12 at `/usr/local/bin/gcc`. This ensures valkey-search builds succeed even on fresh or reset runners without manual setup.

- Checks existing GCC version before installing
- Installs `gcc14`/`gcc14-c++` and creates symlinks (`gcc`, `g++`, `cc`, `c++`)
- No-op on subsequent runs if GCC is already present
- AL2023-specific (uses `dnf`)

### DIFF
--- a/.github/workflows/search_benchmark.yml
+++ b/.github/workflows/search_benchmark.yml
@@ -142,6 +142,35 @@ jobs:
           
           pip install --require-hashes -r requirements.txt
 
+      - name: Ensure GCC >= 12 (required for valkey-search build)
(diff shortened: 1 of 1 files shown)


VoletiRam (approved):

roshkhatri (approved):
I confirmed that out runners have gcc14 / gcc14-c++ repo available good to be merged

PR #81, opened 2026-08-17
Fix server restart failure: flush before shutdown + SIGKILL escalation

VoletiRam:
Fix server restart failure: flush before shutdown + SIGKILL escalation


roshkhatri (approved):
Thanks for addressing the offline feedback. LGTM

PR #83, opened 2026-08-19
Unify basic and test_groups configs into a single execution path

roshkhatri:
`ClientRunner` had two execution trees picked by a `uses_test_groups` flag: one for the basic `commands` config format, one for scenario-based `test_groups`. Scheduling, server lifecycle, population, warmup, CSV parsing and metrics were written twice, so new features either got implemented twice or quietly worked in only one format. Profiling and `--runs` were scenario-only; arbitrary commands were impossible in the basic format.

This merges them. The basic format is compiled into generated `test_groups` at config load and the runner only walks the scenario path. Both formats stay supported and neither changes on the surface, so existing configs and the workflows that edit them keep working.

Scenarios pick up what the basic path used to own privately: `test` for predefined `-t` workloads, plus `data_size`, per-scenario `keyspacelen`, `warmup_inline`, `restart_before` and `populate_with`. `restart_before` and `flush_before` collapsed into one mechanism that also re-applies `config_set` after the restart, which the basic path had been dropping.

## Arbitrary commands

Scenarios using `command` now get the same client settings as predefined ones, which is what makes their numbers comparable: `data_size` reaches `-d`, so `__data__` is the size you asked for rather than valkey-benchmark's 3-byte default, and `benchmark-threads` reaches `--threads`, so the client isn't left single-threaded driving a thousand connections. `populate_with` works for them too.

## Fixes found along the way

Population ran outside the scenario error handler, so a failed populate escaped with no failure marker and silently skipped every later scenario. A `populate_with` that can't run is now rejected at validation instead of being skipped, which used to benchmark reads against an empty keyspace and report the result as if it meant something.

Parallel and mixed execution assumed every scenario had a `command`, so a predefined workload raised `KeyError` or silently produced no metrics. Mixed sub-scenarios also reported a hardcoded `data_size` of 100 regardless of payload. Parallel readers each drew their own seed instead of sharing the populate pass's, so they could read keys that were never written. Profiling cleanup only ran on success paths, so an exception left `perf` running.

Mixed sub-scenarios inherited a duration but not a request count, so a request-bounded mixed run either failed outright for predefined workloads or silently fell back to a 60 second default for arbitrary ones. The count now reaches the sub-scenarios and appears in their metrics rows and failure markers.

Mixed workloads now emit a failure marker per expected sub-scenario, covering partial failure (which previously reported nothing at all), total failure, and a mid-run exception, and orphaned client processes are terminated.

CPU pinning for mixed processes indexes the real core list, so a non-contiguous client pool such as `10-11,20-21` no longer hands a process the absent cores in the gap. Allocation past the end of the pool keeps its existing warn-and-continue behaviour.

## Comparison tool

Several bugs shared one cause: fields that are metadata, not configuration, leaked into the signature used to pair baseline rows with new rows.

`env_`-prefixed metadata is excluded now. This one was live on main: the CPU frequency is read at runtime, so two runs on the same host produced different signatures and the comparison table came out empty. Only genuinely volatile readings are excluded, so two different machines no longer compare as though the software changed.

Signatures are discovered across both datasets, and a `config_set` sweep participates, where distinct server configurations used to be averaged together as repeated runs, quietly turning 100 and 200 into 150. The sweep value now appears in group headings so the tables are attributable.

Failure markers carry the full scenario identity. Without it a failure never paired with its own successful counterpart, so you got a fabricated `-100%` next to an orphan row. Failed scenarios are excluded from the numeric comparison, since they have no metrics, and reported in a table keyed on that identity.

## utils/modify_config.py

Consumer repos build a config by editing one with jq inside their workflow: bounds-checking inputs, deriving `server_cpu_range`, and for arbitrary commands assembling a whole `test_groups` block. That's schema knowledge living outside the repo that owns the schema, it can't be tested there, and every consumer reimplements it.

The script does the same work and validates the result through `validate_config` before writing, so it can't emit a config the runner would reject. Where an input is ambiguous it is rejected rather than quietly resolved: a base carrying both `duration` and `requests`, a single-value field given a list, a command whose name is empty, a client pool too small for the processes a mixed load will start. Validation covers every entry in the file, and CPU range checks read the effective allocation, so the conversion that folds explicit ranges into `cpu_allocation` cannot hide a server and client overlap. It has three modes: predefined commands, one arbitrary command with optional population, and a concurrent read plus write mix. The mixed mode splits the configured client count by a ratio so total load is unchanged, derives the CPU allocation the concurrent processes need, pins the generated scenario to single-node execution so the process count stays fixed rather than multiplying per cluster node, warms with writes only so the warmup budget populates instead of reading a cold keyspace, and takes either a predefined workload or an arbitrary command string on each side. Runner-specific bounds are parameters rather than constants.

It removes about 160 lines from the valkey workflow that calls it.

## Behaviour changes worth calling out

- Compiled basic-format scenarios re-apply `config_set` after restart; those values used to be lost.
- `--groups` and `--scenarios` now filter basic configs; they were accepted and ignored before.
- Comparison output changes as described above.
- Basic-format `metrics.json` field set is unchanged.

## Testing

Parity is pinned by exact-argv tests asserting complete command lines for a compiled write workload, a compiled read workload including its populate pass, and an arbitrary command, plus an integration test covering load, compile and a mocked run.

One lesson is baked in: several of these bugs hid because tests hand-built rows in shapes production never emits, and because comparison tests called `create_comparison_table_data` directly and skipped the averaging stage `main()` runs. Rows are now built by the real producers, and the comparison tests go through averaging.

Full suite: 779 passed, 23 skipped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,6 +317,8 @@ Create benchmark configurations in JSON format. Each object represents a single 
 ]
 ```
 
+Internally, this basic format is compiled into the scenario model at load time: each combination of the list-valued fields (requests x keyspacelen x data_sizes x pipelines x clients x commands) becomes a generated test group containing a single scenario, and the runner executes only the `test_groups` path. This is purely an internal translation: benchmark invocations and `metrics.json` outputs are unchanged, and both the basic format and the `test_groups` format (see Module Config Structure) remain fully supported.
+
 #### Custom Server Configs Examples
 
 Add server configs the benchmark does not manage (e.g. memory limits, timeouts):
@@ -661,6 +663,18 @@ Module tests use structured `test_groups` with `scenarios`:
   "tls_mode": false
 }
 ```
+
+**Scenario fields:** Each scenario names its workload with exactly one of `test` or `command` (except `type: mixed` scenarios, which use `writes`/`reads` sub-scenarios instead):
+
+| Field            | Description                                                                                                                       |
+| ---------------- | --------------------------------------------------------------------------------------------------------------------------------- |
+| `test`           | Predefined valkey-benchmark test name, run as `-t NAME`. Used by compiled basic-format configs.                                    |
+| `command`        | Arbitrary command line, run after `--`. Supports `__rand_int__` placeholders.                                                      |
+| `data_size`      | Payload size in bytes, passed as `-d N` when present.                                                                              |
+| `keyspacelen`    | Per-scenario key space size, passed as `-r N`. Falls back to the config-level `keyspacelen[0]`.                                    |
+| `warmup_inline`  | Adds `--warmup N` to the main benchmark run. Distinct from `warmup`, which performs a separate warm-up run first.                  |
+| `restart_before` | Restarts the managed server before the scenario runs (flushes the database instead when using a running server).                   |
+| `populate_with`  | Write workload that seeds the keyspace before a read test. For a `test` scenario it is a predefined write name (e.g. `SET`), run as `-t NAME`; for a `command` scenario it is an arbitrary write command string (e.g. `SET key:__rand_int__ __data__`), run after `--`. The populate pass runs sequentially and shares the main run's seed, so the read hits the seeded keys. |
 
 ### Cluster Mode Support
 
--- a/benchmark.py
+++ b/benchmark.py
@@ -5,6 +5,7 @@ import argparse
 import json
 import logging
 import platform
+from itertools import product
 from pathlib import Path
 from typing import List, Optional
 import sys
@@ -12,7 +13,14 @@ import sys
 
 from valkey_build import ServerBuilder
 from valkey_server import ServerLauncher, apply_config_to_servers
-from valkey_benchmark import ClientRunner
+from valkey_benchmark import (
+    ClientRunner,
+    ORIGIN_FIELD,
+    ORIGIN_SIMPLE,
+    READ_COMMANDS,
+    READ_POPULATE_MAP,
+    WRITE_COMMANDS,
+)
 from benchmark_build import BenchmarkBuilder
 from utils.cpu_utils import (
     parse_core_range,
@@ -286,6 +294,54 @@ def _validate_cpu_range(value, key_name: str) -> None:
 # ---------- Helpers ----------------------------------------------------------
 
 
+def compile_simple_config(cfg: dict) -> None:
+    """Compile the basic 'commands' format into generated test_groups.
+
+    Each parameter combination becomes a single-scenario group, preserving
+    consecutive run ordering. The origin marker preserves the basic metrics
+    schema and shared seed. Cluster-specific filtering happens later, after
+    cluster mode has been scalarized.
+    """
+    requests_list = (
+        cfg["requests"] if cfg.get("requests") is not None else [None]
+    )  # duration mode has no requests
+
+    groups = []
+    for requests, keyspacelen, data_size, pipeline, clients, command in product(
+        requests_list,
+        cfg["keyspacelen"],
+        cfg["data_sizes"],
+        cfg["pipelines"],
+        cfg["clients"],
+        cfg["commands"],
+    ):
+        if command not in READ_COMMANDS + WRITE_COMMANDS:
+            logging.warning(f"Unsupported command: {command}, skipping.")
+            continue
+
+        scenario = {
+            ORIGIN_FIELD: ORIGIN_SIMPLE,
+            "id": command,
+            "test": command,
+            "data_size": data_size,
+            "pipeline": pipeline,
+            "clients": clients,
+            "keyspacelen": keyspacelen,
+            "warmup_inline": cfg["warmup"],
+            "restart_before": True,
+        }
+        if cfg.get("duration") is not None:
+            scenario["duration"] = cfg["duration"]
+        else:
+            scenario["requests"] = requests
+        if command in READ_COMMANDS:
+            scenario["populate_with"] = READ_POPULATE_MAP[command]
+
+        groups.append({"group": len(groups) + 1, "scenarios": [scenario]})
+
+    cfg["test_groups"] = groups
+
+
 def validate_config(cfg: dict) -> None:
     """Validate config (commands or test_groups format)."""
     if "scenarios" in cfg and "test_groups" not in cfg:
@@ -369,6 +425,9 @@ def validate_config(cfg: dict) -> None:
         cfg["cluster_mode"] = parse_bool(cfg["cluster_mode"])
     if "tls_mode" in cfg:
         cfg["tls_mode"] = parse_bool(cfg["tls_mode"])
+
+    if has_commands and not has_test_groups:
+        compile_simple_config(cfg)
 
 
 def load_configs(path: str) -> List[dict]:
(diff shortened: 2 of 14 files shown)


rainsupreme (approved):
The unification is a good move! I'm approving, assuming you take a look at the notes below and all CI jobs pass. :)

I did have my AI take a look at the PR - here's the notes:

1. **Suggestion: one test for `--groups`/`--scenarios` on basic-format configs.**
   It's a nice behavior change (they used to be silently ignored), and it's exactly
   the kind of thing that could quietly regress in a future refactor since nothing
   pins it today. Cheap insurance for a filter we'll all lean on when slicing runs.

2. **Heads-up for coordination**: #86 (iteration support for test group scenarios)
   builds on the pre-unification code this PR deletes — probably worth a cross-link
   so @bluayer knows a rebase onto the new single path is coming. The good news is
   iteration support should only need implementing once now.

roshkhatri:
Examples of compare script for failed scenarios and other for success

## No significant changes

No statistically significant changes detected across 0 test(s).

## ⚠️ 2 failed scenario(s)

Excluded from the numeric comparison (a failure has nothing to compare against); the measured value is shown for whichever side succeeded:

| Scenario | Phase | Command | ` baseline @octocat #83 ` | ` [new](https://evil.example) GH-83 ` | Error |
| --- | --- | --- | --- | --- | --- |
| ` unsafe_demo @octocat #83 (note=a\|b; policy=[docs](https://evil.example); tls=True) ` | ` read @valkey-io/core-team ` | `` GET [click me](https://evil.example) GH-83 a@example.com \| `inline` `` | ` 123K rps ` | **FAILED** | ``` boom  @octocat #83 https://evil.example `rm -rf /` ``nested`` ``` |
| ` both_failed ` | ` write ` | ` SET user@example.com ` | **FAILED** | **FAILED** | ` baseline: baseline failed: CVE-2024-1234; new: new failed: valkey-io/valkey-perf-benchmark#83 ` |

<details>
<summary>Click to expand full comparison tables</summary>

**Legend:**
- **Test column**: Command, metric, P=pipeline depth, T=io-threads
- **Significance**: ✅ significant improvement, ❌ significant regression, ➖ not significant, ❔ insufficient data

**Statistical Notes:**
- **CV**: Coefficient of Variation - relative variability (σ/μ × 100%)
- **CI95%**: 95% Confidence Interval - range where the true population mean is likely to fall
- **PI95%**: 95% Prediction Interval - range where a single future observation is likely to fall

</details>

---

# Success scenario

# Benchmark Comparison: [baseline123](https://github.com/valkey-io/valkey/commit/baseline123) vs [new456](https://github.com/valkey-io/valkey/commit/new456)

## 5 significant change(s)

- ✅ -17±5% SET p99_latency pipe=1 threads=4 (scenario=write, test_id=1_write, test_phase=write)
- ✅ -15±8% SET p50_latency pipe=1 threads=4 (scenario=write, test_id=1_write, test_phase=write)
- ✅ -14±6% SET p95_latency pipe=1 threads=4 (scenario=write, test_id=1_write, test_phase=write)
- ✅ +12±4% SET rps pipe=1 threads=4 (scenario=write, test_id=1_write, test_phase=write)
- ✅ -12±7% SET avg_latency pipe=1 threads=4 (scenario=write, test_id=1_write, test_phase=write)

*5 with no significant change*

<details>
<summary>Click to expand full comparison tables</summary>

### scenario = read, test_id = 1_read, test_phase = read

| | % Change | Test | [baseline123](https://github.com/valkey-io/valkey/commit/baseline123) | [new456](https://github.com/valkey-io/valkey/commit/new456) | baseline123 stats | new456 stats |
| --- | --- | --- | --- | --- | --- | --- |
| ➖ | +0±4% | GET rps P1 T4 | 200K | 201K | n=3, σ=2.00K, CV=1.0%, CI95%=±2.5%, PI95%=±5.0% | n=3, σ=2.00K, CV=1.0%, CI95%=±2.5%, PI95%=±4.9% |
| ➖ | (+0±1)×10¹% | GET avg_latency P1 T4 | 0.25 | 0.25 | n=3, σ=0.0100, CV=4.0%, CI95%=±9.9%, PI95%=±19.9% | n=3, σ=0.0100, CV=4.0%, CI95%=±9.9%, PI95%=±19.9% |
| ➖ | (+0±1)×10¹% | GET p50_latency P1 T4 | 0.197 | 0.197 | n=3, σ=0.00577, CV=2.9%, CI95%=±7.3%, PI95%=±14.6% | n=3, σ=0.00577, CV=2.9%, CI95%=±7.3%, PI95%=±14.6% |
| ➖ | (+0±1)×10¹% | GET p95_latency P1 T4 | 0.40 | 0.400 | n=3, σ=0.0200, CV=5.0%, CI95%=±12.4%, PI95%=±24.8% | n=3, σ=0.01000, CV=2.5%, CI95%=±6.2%, PI95%=±12.4% |
| ➖ | +0±9% | GET p99_latency P1 T4 | 0.60 | 0.60 | n=3, σ=0.0200, CV=3.3%, CI95%=±8.3%, PI95%=±16.6% | n=3, σ=0.0100, CV=1.7%, CI95%=±4.1%, PI95%=±8.3% |

### scenario = write, test_id = 1_write, test_phase = write

| | % Change | Test | [baseline123](https://github.com/valkey-io/valkey/commit/baseline123) | [new456](https://github.com/valkey-io/valkey/commit/new456) | baseline123 stats | new456 stats |
| --- | --- | --- | --- | --- | --- | --- |
| ✅ | +12±4% | SET rps P1 T4 | 100K | 112K | n=3, σ=1.00K, CV=1.0%, CI95%=±2.5%, PI95%=±5.0% | n=3, σ=1.00K, CV=0.9%, CI95%=±2.2%, PI95%=±4.4% |
| ✅ | -12±7% | SET avg_latency P1 T4 | 0.50 | 0.44 | n=3, σ=0.0100, CV=2.0%, CI95%=±5.0%, PI95%=±9.9% | n=3, σ=0.0100, CV=2.3%, CI95%=±5.6%, PI95%=±11.3% |
| ✅ | -15±8% | SET p50_latency P1 T4 | 0.410 | 0.350 | n=3, σ=0.01000, CV=2.4%, CI95%=±6.1%, PI95%=±12.1% | n=3, σ=0.01000, CV=2.9%, CI95%=±7.1%, PI95%=±14.2% |
| ✅ | -14±6% | SET p95_latency P1 T4 | 0.80 | 0.690 | n=3, σ=0.0200, CV=2.5%, CI95%=±6.2%, PI95%=±12.4% | n=3, σ=0.01000, CV=1.4%, CI95%=±3.6%, PI95%=±7.2% |
| ✅ | -17±5% | SET p99_latency P1 T4 | 1.20 | 1.00 | n=3, σ=0.0200, CV=1.7%, CI95%=±4.1%, PI95%=±8.3% | n=3, σ=0.0200, CV=2.0%, CI95%=±5.0%, PI95%=±9.9% |

---

**Configuration:**
- benchmark_mode: requests
- clients: 50
- cluster_mode: False
- data_size: 64
- group: 1
- requests: 100000
- status: success
- tls: False

**Legend:**
- **Test column**: Command, metric, P=pipeline depth, T=io-threads
- **Significance**: ✅ significant improvement, ❌ significant regression, ➖ not significant, ❔ insufficient data

**Statistical Notes:**
- **CV**: Coefficient of Variation - relative variability (σ/μ × 100%)
- **CI95%**: 95% Confidence Interval - range where the true population mean is likely to fall
- **PI95%**: 95% Prediction Interval - range where a single future observation is likely to fall

</details>

PR #90, opened 2026-09-17
Add benchmark_args scenario key for valkey-benchmark flag passthrough

roshkhatri:
Adds a `benchmark_args` scenario config key: a list of strings passed through to `valkey-benchmark` itself, spliced into the argv before `--csv` and therefore always left of the `--` separator. Each string is shlex-split.

This lets configs use client flags the framework does not model (for example `--zipfian` and `--keysize` from a patched `valkey-benchmark` supplied via `--valkey-benchmark-path`) without a framework change per flag. It targets the benchmark client, unlike `custom-server-configs`, which targets `valkey-server`.

- On a `type: mixed` parent, the value is inherited by every `writes`/`reads` child that does not set its own
- Validated as a list of strings, with a clear error naming the offending scenario
- A scenario without the key builds a byte-identical argv to before

Testing: 16 new tests covering argv placement relative to `--`, shlex splitting, mixed inheritance and child override, unchanged behavior without the key, and validation rejects.


rainsupreme (commented):
This looks like a solid addition to me. My AI thinks it found a couple issues though:

First, mixed children aren't validated, and the failure mode is garbage argv. validate_test_groups checks the parent then continues on type: mixed; nothing iterates writes/reads. So a child with "benchmark_args": "--zipfian 1.0" (string, not list) passes validation, then for arg in scenario.get("benchmark_args", []) iterates characters and shlex.split each one. I confirmed the result: ['-', '-', 'z', 'i', 'p', 'f', 'i', 'a', 'n', '1', '.', '0'] appended to the valkey-benchmark argv, and the command becomes `valkey-benchmark ... --seed 42 - - z i p f i a n 1 . 0 --csv`

Also, these args are emitted after -c, -P, -d, -n/--duration, -t, --seed, and valkey-benchmark is last-wins, so ["-c 50"] runs 50 clients while metrics.json says clients: 1. Same class of problem as port in #93.

It'd be good to add regression tests too ;)

roshkhatri:
>Also, these args are emitted after -c, -P, -d, -n/--duration, -t, --seed, and valkey-benchmark is last-wins, so ["-c 50"] runs 50 clients while metrics.json says clients: 1. Same class of problem as port in https://github.com/valkey-io/valkey-perf-benchmark/pull/93.

Yeah we can add protected args for this as well I think.

PR #91, opened 2026-09-17
Add per-second metrics sampler and tiering time series table

roshkhatri:
Adds `metrics_sampler.py`, a 1 Hz sampler that reads the server and host once per second and emits one row per sample, for per-commit deep-dive dashboards. Each row carries:

- `elapsed_sec` since the measured phase started, so multiple commits overlay on one chart axis
- Tiering `INFO` counters under their own field names, plus `completion_read_ok` and `dram_value_hits`
- DRAM and disk hit percentages in cumulative form (matches the reference dashboard CSV) and per-interval form (shows transients a cumulative ratio dilutes)
- Memory, keyspace hits/misses, throughput from `total_commands_processed` deltas
- Per-thread Valkey CPU with the async IO worker isolated by thread name
- Block device IOPS and MB/s on an auto-detected NVMe or SCSI device

Sampling runs on a background daemon thread and never raises into the caller. A server without tiering samples cleanly with tiering columns at 0.

Sampled rows land in `results_dir/timeseries.json`, appended across scenarios, repeated runs and config sets, with run identity (commit, scenario, command, config_set) on every row. The PostgreSQL table and the push step arrive in a follow-up PR.

Nothing in this PR touches a database: the sampler has no DB code and no workflow pushes its output.

The sampler is wired into the run behind an opt-in `per_second_sampling` config key (default off: the sampler is never constructed and the argv and behavior are unchanged). A `sampler-smoke.yml` CI workflow builds valkey unstable on a GitHub runner, runs `configs/sampler-smoke.json` against the live server, and asserts on the emitted time series: rows present, `elapsed_sec` increasing from 0, real memory and throughput values, tiering counters all 0 on a stock server, and run identity on every row.

Testing: 89 new tests (sampler behavior, ratio math in both forms, zero denominators, missing devices, and the CI verifier script itself).


valkey_benchmark.py
```diff
@@ -985,6 +1006,157 @@ def _stop_scenario_profiling(
...
+        if self.io_threads is not None:
+            context["io_threads"] = self.io_threads
+        if self.architecture is not None:
+            context["architecture"] = self.architecture
+
+        return MetricsSampler(
+            host=self.target_ip,
+            port=self._get_active_ports()[0],
+            cli_path=str(self.valkey_path / VALKEY_CLI),
+            server_pid=self._resolve_server_pid(),
```
rainsupreme, line 1097: The remote-server case reads local `/proc/<remote pid>/` and local disk stats, mis-attributing CPU. Needs a loopback guard + test. With --use-running-server against another host that pid is either absent (silent zeros, fine) or belongs to an unrelated local process, whose CPU is then reported as valkey_cpu_pct. Disk stats have the same locality assumption.

metrics_sampler.py
```diff
@@ -0,0 +1,827 @@
...
+        self._stop_event.set()
+        if self._sampler_thread is not None:
+            self._sampler_thread.join(timeout=_JOIN_TIMEOUT_SEC)
+            if self._sampler_thread.is_alive():
+                logging.warning("Metrics sampler thread did not stop within timeout")
+            self._sampler_thread = None
+
+        logging.info(f"Stopped metrics sampler after {len(self._rows)} samples")
+
+    def write(self, path: Any) -> None:
```
rainsupreme, line 484: looks like a dead function with no callers

rainsupreme (commented):
This will be a nice feature! Seems like it might not work for remote servers though?

xdk-amz:
Overall suggestion:

When I wrote the script for DT specifically, I parsed out the specific fields I wanted to record like you do here. I eventually found it useful to retroactively view INFO fields I hadn't specifically parsed out so I updated the script to dump full INFO output every tick as well. 

For a general purpose metrics sampler in this repo, you should take a similar approach and dump full INFO every tick and then later stages in the pipeline can define what they want to parse.

I'd take it a step further and make `metrics_sampler.py` agnostic to any specific sampling strategy and create an interface that samplers can implement and register with `metrics_sampler.py` as part of scenario configuration. 

The metrics sampler can then loop through each sampler at a fixed rate (1hz, as today) or eventually each sampler can register its own sampling rate (I'd defer this for now). 

But it will be nice to separate different samplers like
```
valkey_info_sampler.py # samples Valkey INFO
disk_stat_sampler.py # samples host disk stats
...
etc.
```
and a scenario can define which it input sources it cares to sample and later stages can define the filtering rules for ingesting those samples



With the current design, adding new metrics to a dashboard requires code changes to metrics_sampler and rerunning the tests. With the proposed design, it can be possible to just modify a dashboard and see those metrics for past and future runs.

PR #93, opened 2026-09-17
Let custom-server-configs override non-protected server defaults

roshkhatri:
A framework server default is now skipped when the config supplies the same key through `custom-server-configs`, so the user's value is the only occurrence and Valkey honors it. Six harness plumbing keys are protected and always win regardless: `cluster-enabled`, `daemonize`, `logfile`, `save`, `appendonly`, `protected-mode`.

Motivation: `maxmemory-policy` was previously overridden to `allkeys-lru` silently, which made configs needing `noeviction` (no-tiering baselines, manual-spill latency scenarios) inexpressible. Note Valkey's own default is `noeviction`; `allkeys-lru` remains this framework's default when a config does not set the key.

A config that sets no colliding key builds a byte-identical command line, verified across 21 config and call-shape combinations. No config in the repo currently sets a colliding key.

Testing: 9 precedence tests (user value wins for a non-protected key, every protected key still wins, no-collision argv unchanged verbatim).


README.md
```diff
@@ -358,15 +358,17 @@ Combine with a baseline `.conf` file:
...
+1. `custom-server-config-file` (positional, parsed first, lowest priority)
+2. Benchmark-managed defaults (`--key value` flags)
+3. `custom-server-configs` (`--key value` flags, highest priority)
 
-This means harness-critical settings (`port`, `daemonize`, `logfile`, `cluster-*`, etc.) always take effect regardless of what the user supplies. Setting them in `custom-server-configs` is allowed but has no effect.
+A key you set in `custom-server-configs` wins over the benchmark default for that key: the framework skips its own default instead of appending it afterwards. For example, setting `"maxmemory-policy": "noeviction"` takes effect even though the default is `allkeys-lru`.
+
+The exception is a small protected set the harness always controls, because the framework depends on it for process management and log capture: `cluster-enabled`, `daemonize`, `logfile`, `save`, `appendonly`, `protected-mode`. Setting one of these in `custom-server-configs` is allowed but has no effect. The same is true of flags the framework sets from other config fields (`port`, `bind`, `io-threads`, `cluster-config-file`, `loadmodule`, TLS paths).
```
rainsupreme, line 371: The described behavior is the right one, but the code doesn't do it: those flags are emitted before the custom block in _build_server_command, so last-wins gives the user's value precedence, the reverse of the protected set. On this branch {"port": "7000", "io-threads": "9"} yields --port 6379 --io-threads 4 ... --port 7000 --io-threads 9: the server binds 7000 while the harness health-checks 6379, and a mismatched io-threads runs but records the wrong value in the results. Pre-existing ordering, but this PR documents it as safe, so let's make it true here.

Two suggested options: reject these keys in the custom-server-configs validator in benchmark.py with a pointer to the first-class field (port, io_threads, tls_mode, modules), or move the custom block up to right after the positional config file so every harness flag lands after it.

Either way, it'd be good to add one test with port in custom-server-configs.

rainsupreme (commented):
Good idea and mostly looks good, but it looks like harness-critical settings might not actually be protected as intended. (Preexisting, but contradicts the readme edit)

Everything else looks good tho :)

